PR #24, opened 2026-04-17
Fix line feeds in conditions

### LINKED ISSUE #23: Condition strings don't support line feeds
Often times it's useful to break the condition into multiple lines. However, the parser you use for the condition, really hates it. You probably need to adjust the grammar to treat line feeds as white space. For example:

```
  condition: >-
    selection and replication_guid
    and not (filter_machine_account or filter_known_service_accounts
             or filter_msol_prefix or filter_nt_authority_prefix)
```

Results in:

```
  |
1 | selection and replication_guid and not (filter_machine_account or filter_known_service_accounts␊
  |                                                                                                ^---
  |
  = expected or_op or and_op
```

fwosar:
Adds line feeds to the grammar as valid whitespace. Also added a unit test to make sure conditions containing line feeds are parsed correctly. Fixes #23.

### DIFF
--- a/crates/rsigma-parser/src/parser.rs
+++ b/crates/rsigma-parser/src/parser.rs
@@ -1825,4 +1825,39 @@ level: low
             Some("process_creation".to_string())
         );
     }
+
+    #[test]
+    fn test_line_feed_in_condition() {
+        let yaml = r#"
+title: Line Feed Condition rule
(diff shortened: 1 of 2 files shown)


mostafa (approved):
@fwosar Thanks for your contribution! LGTM.

PR #26, opened 2026-04-17
Propagate custom rule attributes through results

### LINKED ISSUE #20: Add support for custom Sigma rule fields
Would it be possible to expose any fields that weren't consumed by the parser in matches? Many rules contain custom information that is stored in non-standard fields. The most common example of this are template strings used to pretty print rule matches. Currently, all these fields get dropped as far as I can tell and are not returned as part of a rule match. While it is possible to just read those custom fields yourself based on the returned rule id, it would be way more convenient if these custom fields would just be returned as part of the match in a `custom` property, for example.

fwosar:
I decided not to piggy-back off the existing custom_attributes property. Instead, custom attributes set in rules are propagated by the custom_rule_attributes properties. The reason being is that custom_rule_attributes really wants to be a HashMap pointing to Serde values (as the rule attributes can be nested). Also, it avoids any sort of name conflicts. 

I tried to stick to the existing code style. That being said, I am not a super experienced Rust developer. So if there are more idiomatic ways to do things, please let me know and I can change it.

Closes #20.


mostafa (commented):
Please see https://github.com/timescale/rsigma/issues/20#issuecomment-4270503284 first.

## Overall

The split between the existing typed `custom_attributes` (rsigma.* engine controls, `HashMap<String, String>`) and the new `custom_rule_attributes` (arbitrary user metadata, nested values) is the right call -- widening `custom_attributes` to nested values would have broken engine code that calls `.parse()` on the strings. Two semantically-distinct dicts is cleaner. The detection-rule key list is complete and accurate.

A few items inline to address before merge, plus two design follow-ups worth highlighting.

## Required (see inline comments)

- **Bug**: `standard_correlation_keys` is out of sync with `parse_correlation_rule` -- silent duplication of `name`/`tags`, silent data loss for `taxonomy`/`falsepositives`/top-level `generate`.
- **Visibility**: `yaml_to_json_map` should be `pub(crate)`.
- **Robustness**: NaN/Inf would panic in the `f64` arm of `yaml_to_json`.

## Worth elevating (not blocking -- open as follow-up issues if you agree)

- **Sync-drift hazard between `standard_*_keys` and the struct fields/parser.** The lists live in a different file from the struct definitions and from the `get_str(...)` calls that consume the keys. Adding a new top-level field means updating both, easy to forget. A cleaner pattern would be to invert the predicate: track which keys the parser actually consumed and collect the rest, mirroring serde's `#[serde(flatten)]` behavior. Out of scope for this PR but worth queuing.
- **Per-match clone of the attributes map.** `MatchResult` and `CorrelationResult` deep-clone the whole map on every successful match. With Level-3 batch evaluation (parallel rayon, millions of events/sec) and rules that carry nested attrs, this isn't free. Storing `CompiledRule.custom_rule_attributes` as `Arc<HashMap<String, serde_json::Value>>` and cloning the `Arc` would make this a pointer bump with no public-API churn. Suggest a follow-up PR rather than scope-creep here.

Thanks for splitting the new field cleanly from `custom_attributes` and for the thorough parser tests.

crates/rsigma-eval/src/compiler.rs
```diff
@@ -847,6 +854,50 @@ fn pattern_matches(pattern: &str, name: &str) -> bool {
...
+                .filter_map(|(k, v)| Some((k.as_str()?.to_string(), yaml_to_json(v))))
+                .collect();
+            serde_json::Value::Object(obj)
+        }
+        serde_yaml::Value::Tagged(tagged) => yaml_to_json(&tagged.value),
+    }
+}
+
+/// Convert a map of YAML values to a map of JSON values.
+pub fn yaml_to_json_map(
```
mostafa: Suggest `pub(crate)` -- only `compile_correlation` calls this (via `crate::compiler::yaml_to_json_map`). `pub` adds it to `rsigma-eval`'s public API surface unnecessarily.

```suggestion
pub(crate) fn yaml_to_json_map(
```

crates/rsigma-eval/src/compiler.rs
```diff
@@ -847,6 +854,50 @@ fn pattern_matches(pattern: &str, name: &str) -> bool {
...
+    match value {
+        serde_yaml::Value::Null => serde_json::Value::Null,
+        serde_yaml::Value::Bool(b) => serde_json::Value::Bool(*b),
+        serde_yaml::Value::Number(n) => {
+            if let Some(i) = n.as_i64() {
+                serde_json::Value::Number(i.into())
+            } else if let Some(u) = n.as_u64() {
+                serde_json::Value::Number(u.into())
+            } else if let Some(f) = n.as_f64() {
+                serde_json::json!(f)
```
mostafa: **Robustness**: `serde_json::Number` cannot represent NaN/Infinity, and `serde_json::json!(f64::NAN)` **panics** at runtime. Safer and matches the defensive pattern of the surrounding arms:

```suggestion
                serde_json::Number::from_f64(f)
                    .map(serde_json::Value::Number)
                    .unwrap_or(serde_json::Value::Null)
```

crates/rsigma-eval/src/compiler.rs
```diff
@@ -265,6 +271,7 @@ pub fn evaluate_rule(rule: &CompiledRule, event: &Event) -> Option<MatchResult>
                 matched_selections,
                 matched_fields,
                 event: event_data,
+                custom_rule_attributes: rule.custom_rule_attributes.clone(),
```
mostafa, line 276: **Performance -- per-match clone (see top-level review).** This deep-clones the entire attribute map on every successful match. Recommend a follow-up PR that changes `CompiledRule.custom_rule_attributes` to `Arc<HashMap<String, serde_json::Value>>` so this becomes a pointer bump. Same treatment for the two sites in `correlation_engine.rs`.

Not blocking this PR -- happy to take it on as a follow-up if you'd rather keep this one focused.

crates/rsigma-eval/src/correlation_engine.rs
```diff
@@ -954,6 +958,7 @@ impl CorrelationEngine {
                     timespan_secs: timespan,
                     events,
                     event_refs,
+                    custom_rule_attributes: corr.custom_rule_attributes.clone(),
```
mostafa, line 963: Same per-match clone concern as in `evaluate_rule` (compiler.rs) -- see the top-level review for the suggested `Arc<HashMap<...>>` follow-up. The second clone site ~100 lines below would get the same treatment.

crates/rsigma-eval/src/correlation_engine.rs
```diff
@@ -242,6 +242,10 @@ pub struct CorrelationResult {
     /// Contains up to `max_correlation_events` timestamp + optional ID pairs.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub event_refs: Option<Vec<EventRef>>,
+    /// Custom rule attributes from the original Sigma correlation rule YAML.
+    /// Contains any non-standard top-level fields from the rule definition.
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    pub custom_rule_attributes: HashMap<String, serde_json::Value>,
```
mostafa: **Test gap**: there's no test in this file that triggers a correlation and asserts the resulting `CorrelationResult.custom_rule_attributes` is populated. The compiler-level test in `correlation.rs` covers the compile step but not the runtime emission. Worth a small `process_*` test before merge.

crates/rsigma-parser/src/parser.rs
```diff
@@ -565,6 +601,36 @@ fn parse_correlation_rule(value: &Value) -> Result<CorrelationRule> {
...
+        "references",
+        "date",
+        "modified",
+        "taxonomy",
+        "correlation",
+        "falsepositives",
+        "level",
+        "generate",
+        "custom_attributes",
+    ];
```
mostafa, line 642: **Bug** -- this list is out of sync with what `parse_correlation_rule` actually consumes:

| Key | Issue |
|---|---|
| `name`, `tags` | **missing** but parsed at top-level (`get_str(m, "name")` / `get_str_list(m, "tags")` later in this function); values get **duplicated** between the typed struct field and `custom_rule_attributes` |
| `taxonomy`, `falsepositives` | included here but `CorrelationRule` has no such fields; values are **silently dropped** instead of being propagated as custom attrs |
| `generate` | parsed only from **inside** the `correlation:` mapping (`corr.get(val_key("generate"))`), not at top level; top-level `generate:` is silently dropped |

```suggestion
    let standard_correlation_keys: &[&str] = &[
        "title",
        "id",
        "name",
        "status",
        "description",
        "author",
        "references",
        "date",
        "modified",
        "tags",
        "level",
        "correlation",       // entire nested block
        "custom_attributes", // explicitly read at top level
    ];
```

Side note: see the top-level review for the broader sync-drift design discussion.

fwosar:
> @fwosar I am wondering if it'd be better to merge the two to stay pySigma-aligned and what the blast radius is! Merging them is a breaking change, which I am okay with since the project is new and is getting traction, so it'd be okay to just experiment. WDYT?

It's certainly possible to merge them. The main reason I didn't is mostly that it looked like a more complicated change and I tried to keep my changes as small as possible. But let me address your review feedback and also try to merge the two properties into a single one.

The question still is how conflicts should be handled. What if both the pipeline and the rule attempt to add the same custom attribute?

mostafa:
> The question still is how conflicts should be handled. What if both the pipeline and the rule attempt to add the same custom attribute?

The order of precedence would be like this:
1. Rule YAML top-level `custom_attributes`: block (correlation rules have it today).
2. Rule YAML non-standard top-level fields (user metadata like `template`, `severity_score`, etc.).
3. Pipeline `SetCustomAttribute` transformations, applied in pipeline order.

Each step can overwrite keys set by an earlier step. That gives us three useful invariants:
- Pipelines beat rules.
- Within a rule, explicit beats implicit.
- Simple to explain in one sentence.

We can then warn on overwrite with something like this:

```rust
if custom_attributes.insert(key.clone(), new_value).is_some() {
    log::warn!(
        "custom attribute '{key}' overwritten by {source}; previous value discarded"
    );
}
```

**Possible edge case:** if the user uses an unknown key under `rsigma.*`, which are semi-reserved for the engine controls, we can either accept it (less favorable - conventional) or warn on the unknown keys and then ignore it by emitting `log::warn!("unknown rsigma.* key '{k}'; ignoring")`.

So: defined order + last-write-wins + log warning.

crates/rsigma-parser/src/parser.rs
fwosar: That list was taken straight from the correlation rule schema: https://sigmahq.io/sigma-specification/json-schema/sigma-correlation-rules-schema.json. So the parse_correlation_rule function is wrong.

crates/rsigma-eval/src/compiler.rs
fwosar: Resolved.

crates/rsigma-eval/src/compiler.rs
fwosar: Resolved.

crates/rsigma-eval/src/compiler.rs
fwosar: Resolved.

crates/rsigma-eval/src/correlation_engine.rs
fwosar: Resolved.

crates/rsigma-eval/src/correlation_engine.rs
fwosar: Resolved.

fwosar:
Merging them certainly would be the cleanest approach and as long as precedence is well-defined, it's certainly a good idea.

I also agree that to avoid the risk of drift, having the parse_* function move everything that wasn't consumed into the custom_attributes might be a cleaner approach compared to the approach I have taken. However, it appears that the parse_correlation_rule function in particular isn't standard compliant, as what it consumes differs from the official schemas.

mostafa:
@fwosar Thank you for your contribution. I'll take it from here.

crates/rsigma-parser/src/parser.rs
mostafa: Good catch! I'll fix it.

mostafa (approved):

PR #27, opened 2026-04-17
Unify `custom_rule_attributes` into `custom_attributes`

mostafa:
Merges the two per-rule dictionaries into a single `custom_attributes` field on `SigmaRule` and `CorrelationRule`, typed as `HashMap<String, serde_yaml::Value>` to preserve nested structures and non-string values. Mirrors pySigma's single `custom_attributes` dict.

Sources, applied in order (last-write-wins):

1. Arbitrary top-level YAML keys not in the standard Sigma schema.
2. Entries of the optional top-level `custom_attributes:` mapping.
3. Pipeline `SetCustomAttribute` transformations, applied after parsing.

Follow-on changes:
- `MatchResult` and `CorrelationResult` expose a single `custom_attributes: Arc<HashMap<String, serde_json::Value>>` (renamed from `custom_rule_attributes`).
- `apply_custom_attributes` in the correlation engine and all `rsigma.*` readers now look up `serde_yaml::Value::as_str()`.
- Pipeline `SetCustomAttribute` stores `Value::String(...)`.
- Parser tests cover the explicit-block override of arbitrary keys; compiler tests cover pipeline overriding the rule YAML.

BREAKING: `custom_rule_attributes` is removed from the public AST and runtime types. Code reading `rule.custom_attributes` must now treat values as `serde_yaml::Value` (use `.as_str()` to recover the old string view). Since #26 was merged recently and not released, this is okay.

xref https://github.com/timescale/rsigma/issues/20#issuecomment-4270503284

@fwosar I'd appreciate your review! 🙏

### DIFF
--- a/crates/rsigma-eval/src/compiler.rs
+++ b/crates/rsigma-eval/src/compiler.rs
@@ -44,10 +44,11 @@ pub struct CompiledRule {
     /// Whether to include the full event JSON in the match result.
     /// Controlled by the `rsigma.include_event` custom attribute.
     pub include_event: bool,
-    /// Custom rule attributes from the original Sigma rule YAML.
-    /// Non-standard top-level fields are propagated to match results.
-    /// Wrapped in `Arc` so that per-match cloning is a pointer bump.
-    pub custom_rule_attributes: Arc<HashMap<String, serde_json::Value>>,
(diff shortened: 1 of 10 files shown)


fwosar:
Looks good to me :)

PR #40, opened 2026-04-26
  feat: support one-to-many in field_name_mapping

fwosar:
Match pySigma's FieldMappingTransformation: a single source field name in `field_name_mapping` can now map to a list of alternative target field names. When more than one alternative is supplied, the matched detection item is replaced with an OR-conjunction over the alternatives. If the surrounding `AllOf` selection has other items, they're preserved across each branch via a Cartesian expansion so AND/OR semantics stay correct.

Wire-compatible at the YAML layer: existing `mapping: {a: b}` pipelines continue to parse: `parse_string_or_list_mapping` accepts both string and list values per key. `Transformation::FieldNameMapping`'s value type changes from `HashMap<String, String>` to `HashMap<String, Vec<String>>`, which is a source-level break for any direct in-Rust constructors.

Correlation rules consume only the first listed alternative for `group_by`, `aliases` mapping values, and threshold `field` as those positions are inherently scalar and OR semantics aren't expressible there. Other field-name transforms (`field_name_prefix`, `field_name_suffix`, `field_name_prefix_mapping`, `field_name_transform`) remain conceptually 1:1; they wrap their single result internally.

Tests: 3 new unit tests covering AnyOf expansion, Cartesian product when multiple items expand, and the single-alternative fast path; 2 new end-to-end YAML integration tests.

### DIFF
--- a/crates/rsigma-eval/src/pipeline/transformations.rs
+++ b/crates/rsigma-eval/src/pipeline/transformations.rs
@@ -25,7 +25,20 @@ use crate::error::{EvalError, Result};
 #[derive(Debug, Clone)]
 pub enum Transformation {
     /// Map field names via a lookup table.
-    FieldNameMapping { mapping: HashMap<String, String> },
+    ///
+    /// Supports pySigma-compatible one-to-many mapping: a single source name
+    /// can map to a list of alternative field names. When more than one
+    /// alternative is present, the matched detection item is replaced with
+    /// an OR-conjunction (`AnyOf`) of items, one per alternative — preserving
+    /// the rule's original AND structure across the rest of the items in the
+    /// same selection via a Cartesian expansion.
+    ///
+    /// Correlation rules use only the first alternative for `group_by`,
+    /// `aliases` mapping values, and threshold `field` (these positions are
+    /// inherently scalar; OR semantics aren't expressible there).
+    FieldNameMapping {
+        mapping: HashMap<String, Vec<String>>,
+    },
 
     /// Map field name prefixes.
     FieldNamePrefixMapping { mapping: HashMap<String, String> },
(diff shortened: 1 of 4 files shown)


crates/rsigma-eval/src/pipeline/transformations.rs
```diff
@@ -528,6 +590,27 @@ fn transform_detection_fields<F>(
...
+            for elem in &group {
+                let mut combo = prefix.clone();
+                combo.push(elem.clone());
+                next.push(combo);
+            }
+        }
+        result = next;
+    }
+    result
+}
```
mostafa, line 643: This function has no size guard. If a rule's `AllOf` has, say, 10 detection items and each maps to 5 alternatives, that is 5^10 = ~9.7 million branches. This would silently balloon memory and CPU. pySigma guards against this in practice because it typically processes rules for query generation (one-shot), but rsigma evaluates live events, so a blown-up detection tree sits in the hot path permanently.

My recommendation: add an upper bound (like 1024 or 4096 combinations, maybe configurable?) and return an error or emit a warning when exceeded. Even a `tracing::warn!` would help users debug silent OOM.

mostafa (commented):
Thanks for your contribution! 🙏 

The only blockers to merging this PR are:
1. Formatting of the code.
2. Upper bound warning for the `cartesian_product` function.

crates/rsigma-eval/src/pipeline/transformations.rs
fwosar: I added a limit as a hardcoded constant for now. I also looked into what it would take to make it user configurable. The cleanest way would probably be to implement an EngineConfig object that gets passed into the Engine and CorrelationConfig would contain a new EngineConfig object as well. The config object would then be handed down the entire chain to the transformation. It would be helpful if you expect there to be more config knobs in the future.

mostafa (approved):

fwosar:
Signed the commit now to meet your merge requirements. :)

mostafa (approved):

PR #161, opened 2026-06-02
feat(lint): add --tag-namespace flag and config file support

fwosar:
Adds a --tag-namespace flag to rsigma rule lint (repeatable) and a tag_namespaces list field to .rsigma-lint.yml so teams can register organisation-specific tag namespaces without disabling the unknown_tag_namespace rule entirely.

On the library side, LintConfig gains a tag_namespaces field that is merged through the same layering as disabled_rules and exclude_patterns. lint_yaml_str_with_config now passes the extra namespaces down to lint_detection_rule at check time rather than filtering after the fact, so the warning message lists the full combined set of known namespaces when it does fire.

Two new tests cover both paths: extra namespaces passed through LintConfig suppress the warning, and the tag_namespaces key in a config file round-trips correctly through LintConfig::load.

### DIFF
--- a/crates/rsigma-parser/src/lint/mod.rs
+++ b/crates/rsigma-parser/src/lint/mod.rs
@@ -535,8 +535,7 @@ pub fn lint_yaml_value(value: &Value) -> Vec<LintWarning> {
     lint_yaml_value_ext(value, &[])
 }
 
-/// Lint a raw YAML string, returning warnings with resolved source spans.
-pub fn lint_yaml_str(text: &str) -> Vec<LintWarning> {
+fn lint_yaml_str_ext(text: &str, extra_ns: &[String]) -> Vec<LintWarning> {
     let mut all_warnings = Vec::new();
 
     for doc in yaml_serde::Deserializer::from_str(text) {
@@ -561,14 +560,18 @@ pub fn lint_yaml_str(text: &str) -> Vec<LintWarning> {
             }
         };
 
-        let warnings = lint_yaml_value(&value);
-        for mut w in warnings {
+        for mut w in lint_yaml_value_ext(&value, extra_ns) {
             w.span = resolve_path_to_span(text, &w.path);
             all_warnings.push(w);
         }
     }
 
     all_warnings
+}
+
+/// Lint a raw YAML string, returning warnings with resolved source spans.
+pub fn lint_yaml_str(text: &str) -> Vec<LintWarning> {
+    lint_yaml_str_ext(text, &[])
 }
 
 fn resolve_path_to_span(text: &str, path: &str) -> Option<Span> {
@@ -798,7 +801,11 @@ impl LintConfig {
             disabled_rules,
             severity_overrides,
             exclude_patterns: raw.exclude,
-            tag_namespaces: raw.tag_namespaces,
+            tag_namespaces: raw
+                .tag_namespaces
+                .into_iter()
+                .map(|s| s.to_lowercase())
+                .collect(),
         })
     }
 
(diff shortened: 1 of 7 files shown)


fwosar:
Sorry, fixed the formatting :)

crates/rsigma-parser/src/lint/mod.rs
```diff
@@ -971,9 +982,38 @@ pub fn apply_suppressions(
...
+                        end_line: loc.line().saturating_sub(1) as u32,
+                        end_col: loc.column() as u32 + 1,
+                    });
+                }
+                all_warnings.push(w);
+                break;
+            }
+        };
+
+        for mut w in lint_yaml_value_ext(&value, &config.tag_namespaces) {
```
mostafa: `lint_yaml_str_with_config` now copies the whole deserialize loop, parse-error span handling, and break-on-error logic from `lint_yaml_str` verbatim, with the only difference being `lint_yaml_value_ext(&value, &config.tag_namespaces)` in place of `lint_yaml_value(&value)`. These two copies will drift over time (for example if the parse-error handling changes). You already introduced the clean `lint_yaml_value` / `lint_yaml_value_ext` split, so please mirror it one level up: extract a private helper like `collect_warnings(text, extra_ns) -> Vec<LintWarning>`, have `lint_yaml_str` call it with `&[]`, and have `lint_yaml_str_with_config` call it with `&config.tag_namespaces` and then apply suppressions.

crates/rsigma-parser/src/lint/rules/detection.rs
```diff
@@ -194,12 +198,18 @@ pub(crate) fn lint_detection_rule(m: &yaml_serde::Mapping, warnings: &mut Vec<Li
                 } else {
                     if let Some(ns) = tag.split('.').next()
                         && !KNOWN_TAG_NAMESPACES.contains(&ns)
+                        && !extra_namespaces.iter().any(|n| n.as_str() == ns)
```
mostafa, line 201: This comparison only runs after `is_valid_tag()` has passed, so `ns` is always lowercase. A namespace registered with uppercase (for example `--tag-namespace MyOrg`) can therefore never match and is silently ignored with no feedback. Consider lowercasing the configured namespaces (or rejecting non-conforming tokens) so the flag behaves predictably.

crates/rsigma-parser/src/lint/mod.rs
```diff
@@ -822,6 +831,8 @@ impl LintConfig {
         }
         self.exclude_patterns
             .extend(other.exclude_patterns.iter().cloned());
+        self.tag_namespaces
```
mostafa, line 841: `tag_namespaces` is a `Vec` and `merge` extends it, so a config-file entry plus a CLI flag (or a repeated flag) can produce duplicates, which then appear twice in the `known namespaces: ...` warning text. Suppression still works, and this matches how `exclude_patterns` already behaves, so it is optional, but de-duping when building `known_list` (or storing the set) would tidy up the message.

crates/rsigma-parser/src/lint/mod.rs
```diff
@@ -1537,6 +1577,7 @@ detection:
                 .into_iter()
                 .collect(),
             exclude_patterns: vec!["test/**".to_string()],
+            ..Default::default()
```
mostafa: The merge test now compiles against the new field via `..Default::default()`, but it does not assert that `tag_namespaces` actually merges. The merge layering is the part most likely to regress, so please add an assertion that base and other namespaces are unioned.

crates/rsigma-parser/src/lint/mod.rs
```diff
@@ -1715,4 +1756,46 @@ EventID: 4624
...
+    }
+
+    #[test]
+    fn extra_tag_namespace_from_config_file() {
+        let yaml = r#"
+tag_namespaces:
+  - myorg
+  - internal
+"#;
+        let tmp = std::env::temp_dir().join("rsigma_test_tag_ns.yml");
```
mostafa: This writes to a fixed path in the system temp dir. It matches the neighbouring `lint_config_exclude_from_yaml`, so it is consistent, but `tempfile` is already a dev-dependency and used elsewhere in this file (`tempfile::tempdir()`). A fixed name can collide across concurrent runs and leaks the file if the test panics before cleanup. Prefer `tempfile` here. Optional.

mostafa (commented):
Thanks for this. The feature is well scoped and works end to end. I pulled the branch and verified the fmt fix lands: `cargo fmt --all -- --check` now passes, `cargo clippy --all-targets --all-features -- -D warnings` is clean on the touched crates, the parser lint tests pass, and `--tag-namespace myorg` correctly suppresses `unknown_tag_namespace`.

A few follow-ups inline. Most are optional, but one is a docs-sync requirement that has no diff anchor so I will note it here: `docs/developers/linter-and-lsp.md` (the Suppressions section) enumerates how CLI flags map to `LintConfig` fields (`--disable` to `disabled_rules`, `--exclude` to `exclude_patterns`, plus `severity_overrides`). Please add the new `--tag-namespace` to `tag_namespaces` mapping there so the developer docs stay in sync with the three user-facing pages you already updated.

crates/rsigma-parser/src/lint/mod.rs
mostafa: We should probably fix `exclude_patterns` as well, but on a follow-up PR.

fwosar:
All your suggestions are implemented now.

mostafa (approved):
Thank you for your contribution! LGTM! I'll raise another PR to fix `exclude_pattern`.

PR #184, opened 2026-06-05
input/syslog: allow parsing syslog messages, where \u{FEFF} is present

sodomelle:


sodomelle:
I wanted it for myself. Sorry.

PR #185, opened 2026-06-05
PR #01-1.0 - Phase 1 Core Foundation

SecurityEnthusiast:
## Summary
- Complete `rstix` Phase 1 core foundation in one focused change.
- Add workspace-integrated STIX core primitives, deterministic SCO ID helpers, and initial vocabulary tables.
- Align docs and fuzz wiring for the new crate surface.
## What is included
- `rstix` crate integrated in workspace with foundational core modules.
- Deterministic ID/timestamp/spec-version/language/confidence primitives.
- Initial vocabulary coverage and supporting documentation updates.
## Test plan


docs/reference/security.md
```diff
@@ -7,12 +7,12 @@ For the SECURITY policy and disclosure process, see [`SECURITY.md`](../security-
...
 
 | Limit | Constant | Value | Scope | Behaviour on overrun |
-|-------|----------|-------|-------|----------------------|
+| ------- | ---------- | ------- | ------- | ---------------------- |
 | Single event size | `MAX_LINE_BYTES` | 1 MiB | `engine daemon` HTTP/stdin event ingest | Line rejected with a `413`-equivalent error in HTTP mode; counted in `rsigma_events_parse_errors_total`. |
 | Condition expression length | `MAX_CONDITION_LEN` | 64 KiB | Rule parser | Rule rejected at parse time with an `InvalidCondition` error. |
 | Condition expression depth | `MAX_CONDITION_DEPTH` | 64 | Rule parser | Same. |
 | JSON event traversal depth | `MAX_NESTING_DEPTH` | 64 | Keyword search inside nested JSON | Traversal stops; deeper fields are not matched against keyword detections. |
-| Windash expansion | `MAX_WINDASH_DASHES` | 8 | `|windash` modifier (5^8 variants) | Compile error if a value contains more than 8 dashes. |
+| Windash expansion | `MAX_WINDASH_DASHES` | 8 | `\|windash` modifier (5^8 variants) | Compile error if a value contains more than 8 dashes. |
```
mostafa:
```suggestion
| Windash expansion | `MAX_WINDASH_DASHES` | 8 | `|windash` modifier (5^8 variants) | Compile error if a value contains more than 8 dashes. |
```

plan/PRs report.md
```diff
@@ -0,0 +1,9 @@
+# PRs report
+
+## PR-001 - Phase 1 Core Foundation
+
+- Branch: `feat/rstix-phase1-core-foundation`
+- Scope: Complete Phase 1 delivery (`core`, `id`, `vocab`) including infrastructure setup, IDs, timestamps, confidence scales, spec version, language tags, query traits, deterministic SCO IDs, vocabulary tables, and docs alignment
+- Commit(s):
+  - included in branch history
+- Status: ready for review
```
mostafa: Remove this file:

```suggestion
```

.gitignore
```diff
@@ -19,3 +21,5 @@ editors/vscode/*.vsix
 .cursor/
 
 /docs-drafts/
+/plan/
```
mostafa, line 24: plan was committed before being ignored.

Cargo.lock
```diff
@@ -5650,6 +5748,19 @@ dependencies = [
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "15053d8d85c7eccdbefef60f06769760a563c7f0a9d6902a13d35c7800b0ad65"
+dependencies = [
+ "futures-util",
+ "js-sys",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "web-sys",
+]
+
```
mostafa: Why do we need this?

crates/rstix/src/id/sco.rs
```diff
@@ -0,0 +1,242 @@
...
+    sco_type: ScoKind,
+    full_value: &serde_json::Value,
+) -> serde_json::Value {
+    let mut result = serde_json::Map::new();
+    match sco_type {
+        ScoKind::Artifact => {
+            if full_value.get("payload_bin").is_some() {
+                clone_field(full_value, "payload_bin", &mut result);
+            } else {
+                clone_field(full_value, "url", &mut result);
```
mostafa: STIX 2.1 ID contributing properties for `artifact` are `hashes, payload_bin` only. This arm adds `url` (not a contributing property) and drops `hashes` when `payload_bin` is present. Both change the UUIDv5, so generated IDs will not match other STIX 2.1 implementations. Include `hashes` and `payload_bin` when present, and only those.

crates/rstix/src/id/sco.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+        }
+        ScoKind::AutonomousSystem => clone_field(full_value, "number", &mut result),
+        ScoKind::Directory => clone_field(full_value, "path", &mut result),
+        ScoKind::DomainName => clone_field(full_value, "value", &mut result),
+        ScoKind::EmailAddr => clone_field(full_value, "value", &mut result),
+        ScoKind::EmailMessage => {
+            clone_field(full_value, "from_ref", &mut result);
+            clone_field(full_value, "subject", &mut result);
+            clone_field(full_value, "date", &mut result);
```
mostafa: STIX 2.1 contributing properties for `email-message` are `from_ref, subject, body`. This uses `date` instead of `body`, which yields non-interoperable IDs.

crates/rstix/src/id/sco.rs
```diff
@@ -0,0 +1,242 @@
...
+        }
+        ScoKind::File => {
```
mostafa, line 69: Several STIX 2.1 deviations for `file`:

1. Contributing properties are `hashes, name, extensions, parent_directory_ref` together. This treats `hashes` as mutually exclusive with the rest, so a file carrying both a hash and a name gets the wrong ID, and `extensions` is never included.
2. The hash preference order on line 74 is `[SHA-256, SHA-1, MD5]`, but the spec requires single-hash selection in the order `[MD5, SHA-1, SHA-256, SHA-512]`. The current order is reversed and omits SHA-512.
3. `pick_preferred_hash` (lines 42-46) returns the entire hashes map when none of the listed keys match, which can emit multiple hashes and violates the include-only-one-hash rule.

crates/rstix/src/id/sco.rs
```diff
@@ -0,0 +1,242 @@
...
+            } else {
+                clone_field(full_value, "name", &mut result);
+                clone_field(full_value, "parent_directory_ref", &mut result);
+            }
+        }
+        ScoKind::Ipv4Addr | ScoKind::Ipv6Addr | ScoKind::MacAddr | ScoKind::Url => {
+            clone_field(full_value, "value", &mut result);
+        }
+        ScoKind::Mutex => clone_field(full_value, "name", &mut result),
+        ScoKind::NetworkTraffic => {
```
mostafa, line 83: STIX 2.1 contributing properties for `network-traffic` are `start, end, src_ref, dst_ref, src_port, dst_port, protocols, extensions`. This omits `start`, `end`, and `extensions`.

crates/rstix/src/id/sco.rs
```diff
@@ -0,0 +1,242 @@
...
+            clone_field(full_value, "dst_ref", &mut result);
+            clone_field(full_value, "src_ref", &mut result);
+            clone_field(full_value, "dst_port", &mut result);
+            clone_field(full_value, "src_port", &mut result);
+            clone_field(full_value, "protocols", &mut result);
+        }
+        ScoKind::Process => {
```
mostafa: Blocker: STIX 2.1 mandates a UUIDv4 for `process` (the spec text: since all properties on this object are optional, please use a UUIDv4 for the ID of this object). Deriving a deterministic UUIDv5 here is a MUST violation; `process` should not go through deterministic generation at all.

crates/rstix/src/id/sco.rs
```diff
@@ -0,0 +1,242 @@
...
+            clone_field(full_value, "pid", &mut result);
+            clone_field(full_value, "command_line", &mut result);
+            clone_field(full_value, "image_ref", &mut result);
+        }
+        ScoKind::Software => {
```
mostafa, line 94: STIX 2.1 contributing properties for `software` are `name, cpe, swid, vendor, version`. This omits `swid`.

crates/rstix/src/id/sco.rs
```diff
@@ -0,0 +1,242 @@
...
+            clone_field(full_value, "account_type", &mut result);
+            clone_field(full_value, "user_id", &mut result);
+            clone_field(full_value, "account_login", &mut result);
+        }
+        ScoKind::WindowsRegistryKey => {
+            clone_field(full_value, "key", &mut result);
+            clone_field(full_value, "values", &mut result);
+        }
+        ScoKind::X509Certificate => {
+            if let Some(hashes) = pick_preferred_hash(full_value, &["SHA-256"]) {
```
mostafa: STIX 2.1 contributing properties for `x509-certificate` are `hashes, serial_number`. Two issues:

1. This adds `issuer` (not a contributing property) and drops `serial_number` when hashes are present; per spec they contribute together.
2. The single-hash selection is wrong. The spec says: if hashes is present, include only one hash, chosen from `[MD5, SHA-1, SHA-256, SHA-512]` in that order of preference. This only checks `SHA-256` (line 113), so an x509 carrying MD5 and SHA-256 picks SHA-256 here but MD5 per spec; and when none of the checked keys are present, `pick_preferred_hash` returns the entire hashes map, which can emit multiple hashes.

crates/rstix/src/id/sco.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// Generate deterministic STIX SCO ID.
+pub fn generate_sco_id(
+    sco_type: ScoKind,
+    full_value: &serde_json::Value,
+) -> Result<StixId, DeterministicIdError> {
+    let selected = select_id_contributing_properties(sco_type, full_value);
+    let canonical = canonicalize_for_id(&selected)
+        .map_err(|err| DeterministicIdError::JcsFailed(err.to_string()))?;
+    let uuid = uuid::Uuid::new_v5(&STIX_SCO_NAMESPACE, &canonical);
```
mostafa, line 142: Blocker: STIX 2.1 section 2.9 requires a UUIDv4 fallback. If the contributing properties are all optional and none are present, a UUIDv4 MUST be used. This always calls `new_v5`, so any SCO that selects to an empty object (for example a `file` with only `size`) canonicalizes to `{}` and every such object collides on one shared ID.

Separately, the `map_err` to `UnknownScoType` on line 138 is unreachable: `ScoKind` is always valid and `StixId::parse` cannot fail for a `{valid-type}--{uuid}` string, so the variant mislabels what would be an internal invariant violation.

Also: the tests only assert determinism and the type prefix, never equality against a known reference UUID. Please add golden vectors from python-stix2 or the spec SCO examples so these conformance issues are actually caught.

crates/rstix/src/core/timestamp.rs
```diff
@@ -0,0 +1,170 @@
...
+
+fn parse_utc(input: &str) -> Result<OffsetDateTime, TimestampError> {
+    if !input.ends_with('Z') {
+        return Err(TimestampError::Invalid(input.to_owned()));
+    }
+    OffsetDateTime::parse(input, &Rfc3339).map_err(|_| TimestampError::Invalid(input.to_owned()))
+}
+
+/// STIX timestamp preserving fractional precision.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
```
mostafa: `StixTimestamp` derives `Eq`/`Ord` over both `inner` and `subsec_digits`, so the same instant with different fractional widths compares unequal and orders by display precision (for example `...12Z` sorts before `...12.000Z`). Since `QueryableStixObject::created`/`modified` feed these into the query layer, comparison and hashing should be by instant only; keep `subsec_digits` as display-only state.

crates/rstix/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+marking = []
+store = []
+enrichment = ["store", "graph"]
+taxii = ["serde", "dep:reqwest", "dep:tokio", "dep:secrecy"]
+testing = ["dep:wiremock"]
+
+full = ["pattern", "validate", "graph", "marking", "store", "enrichment", "taxii"]
+
+[dependencies]
+phf = { version = "0.11", features = ["macros"] }
```
mostafa: `rstix` pins `phf = "0.11"` but the workspace already resolves `phf 0.13.1`, so this adds a second `phf` version to `Cargo.lock`. Bump to `0.13` to dedupe.

crates/rstix/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+serde_jcs = "0.2"
+thiserror = "2"
+time = { version = "0.3", features = ["formatting", "parsing"] }
+uuid = { version = "1", features = ["v4", "v5"] }
+
+serde = { version = "1", features = ["derive"], optional = true }
+serde_json = "1"
+
+reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "json", "stream"], optional = true }
```
mostafa: These optional deps (`reqwest`, `tokio`, `secrecy`, plus `wiremock`) and the nine feature flags are added for modules that are all one-line stubs in this PR, inflating `Cargo.lock` (+111) and `fuzz/Cargo.lock` (+416). This is also the root cause of the `wasm-streams` entry I flagged in `Cargo.lock`: it is transitive from the `reqwest` stream feature (built only for wasm32). Suggest deferring these to the phases that actually use them.

fuzz/fuzz_targets/fuzz_rstix_parse_bundle.rs
```diff
@@ -0,0 +1,9 @@
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    if let Ok(input) = std::str::from_utf8(data) {
+        let _ = rstix::parse_bundle(input);
```
mostafa: This fuzz target only calls `parse_bundle`, which currently returns `NotImplemented`, so it cannot exercise any real code and will just burn CI time. Suggest deferring the target (and its seeds and the CI matrix entry) until `parse_bundle` is implemented.

crates/rstix/README.md
```diff
@@ -0,0 +1,95 @@
...
+let result = parse_bundle(r#"{"type":"bundle","id":"bundle--00000000-0000-0000-0000-000000000000"}"#);
+assert!(matches!(result, Err(ParseError::NotImplemented)));
+
+let id = StixId::parse("indicator--550e8400-e29b-41d4-a716-446655440000").unwrap();
+let typed = IndicatorId::from_stix_id(id).unwrap();
+assert_eq!(typed.as_stix_id().type_name(), "indicator");
+
+let file = serde_json::json!({"hashes":{"SHA-256":"abc"}});
+let selected = select_id_contributing_properties(rstix::core::ScoKind::File, &file);
+let sco_id = generate_sco_id(rstix::core::ScoKind::File, &selected).unwrap();
```
mostafa: `generate_sco_id` already calls `select_id_contributing_properties` internally, so pre-selecting here and passing the result back in double-applies selection. Callers should pass the full SCO object; the example is misleading as written.

crates/rstix/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+
+/// Testing utilities, including mock infrastructure.
+#[cfg(feature = "testing")]
+pub mod testing;
+
+/// Top-level parse error.
+#[non_exhaustive]
+#[derive(Debug, thiserror::Error)]
+pub enum ParseError {
+    /// Returned while Phase 0 infrastructure is being implemented.
```
mostafa: Phase labeling is inconsistent: this says Phase 0, while the PR title and README state Phase 1 is complete (and `docs/reference/architecture.md` also says Phase 0 stub). Pick one label across the crate and docs.

mostafa (commented):
Thanks for kicking off rstix. The scaffolding, typed-ID pattern, and test hygiene are solid, and fmt/clippy/tests all pass locally. Requesting changes on one thing that matters: the deterministic SCO ID logic in id/sco.rs does not conform to STIX 2.1, and the bugs are silent (valid-looking but wrong IDs).

Blockers:
- process must use UUIDv4, not a derived UUIDv5.
- No UUIDv4 fallback when no contributing properties are present, so empty SCOs collide on one ID.
- Wrong/missing contributing properties for file, x509-certificate, email-message (date vs body), network-traffic, artifact, software, plus reversed/incomplete hash selection.
- Tests only assert determinism and the type prefix, never equality against a known UUID. Please add golden vectors from python-stix2 or the spec SCO examples.

Also worth addressing: StixTimestamp Eq/Ord is precision-sensitive, phf duplicates the workspace version, and reqwest/tokio/secrecy plus the fuzz target are added for stub-only modules. Details inline. Happy to re-review once the ID generation is pinned against reference vectors.

mostafa:
The failing jobs are all stuck at cargo's lockfile check, not real code/test failures:

```
error: the lock file Cargo.lock needs to be updated but --locked was passed to prevent this
```

Every `--locked` job (Check, Clippy, Test x3, Coverage, Doc, MSRV, Sigma Corpus) dies instantly on this, while Format, Build docs, and cargo audit pass because they don't use `--locked`.

Root cause: `main` bumped the workspace version `0.13.0` to `0.14.0` after this branch was cut. CI builds the PR auto-merged into `main` (`refs/pull/185/merge`), so `Cargo.toml` now requires `0.14.0` for all workspace crates (via `version.workspace = true`), but the merged `Cargo.lock` still pins the newly added `rstix` at `0.13.0` since `main` never touched it. Lock is out of date, so `--locked` refuses to run. I reproduced this by merging the head into `main` locally and confirmed the only change cargo wants is `rstix` `0.13.0` to `0.14.0`.

Fix: sync with `main` and regenerate the lock.

```bash
git merge origin/main   # or rebase
cargo check --workspace --all-features
git add Cargo.toml Cargo.lock && git commit && git push
```

SecurityEnthusiast:
Thanks for the detailed review — I addressed all items and pushed updates.

### Changes completed

- **Fixed STIX 2.1 SCO ID conformance** in `crates/rstix/src/id/sco.rs`:
  - corrected contributing properties for `artifact`, `email-message`, `file`, `network-traffic`, `software`, and `x509-certificate`
  - corrected hash preference order
  - added UUIDv4 behavior for `process`
  - added UUIDv4 fallback when contributing properties are absent
  - added golden vectors validated against `python-stix2`
- Updated `StixTimestamp` comparison/hashing semantics to be instant-based.
- Bumped `phf` to `0.13` and removed deferred stub-only dependency/feature/fuzz wiring.
- Fixed README/docs consistency points and removed `plan/PRs report.md` from PR scope.

### Validation run locally

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- `cargo test --workspace --all-features --locked`
- `cargo +1.88.0 check --workspace --all-features --locked`
- `cargo llvm-cov --workspace --all-features --locked --lcov --output-path lcov.info`

mostafa (approved):
@SecurityEnthusiast Thank you for your contribution! 🙏

I made some minor changes to make it mergeable.

PR #187, opened 2026-06-08
feat(syslog): strip UTF-8 BOM from RFC 5424 messages

mostafa:
## Summary

RFC 5424 section 6.4 mandates that a UTF-8 `MSG` begin with a byte order mark (`U+FEFF`, bytes `EF BB BF`) as an encoding marker, not as content. `syslog_loose` preserves it verbatim, and `str::trim()` does not remove it (`U+FEFF` is not Unicode `White_Space`), so the BOM leaked into the parsed event. It corrupted the `_raw` field and anchored matchers (`startswith`, exact equality), and it blocked embedded-JSON detection because `serde_json` errors on a leading BOM, silently degrading a BOM-prefixed JSON payload to a `KvEvent`.

- **Runtime**: the syslog adapter now strips a single leading BOM from the message body by default, gated by a new `SyslogConfig.strip_bom` field (manual `Default = true`, so `InputFormat::default()` strips).
- **CLI/config**: a value-taking `--syslog-strip-bom <BOOL>` flag (default true) on `engine eval` and `engine daemon`, plus the matching `input.syslog_strip_bom` / `eval.syslog_strip_bom` config keys. The flag and config key share one positive name to avoid double negatives; the drift-guard test pins the clap default to the config constant.
- **Docs**: input-formats guide subsection, eval/daemon CLI references, both CLI README tables, and a CHANGELOG entry.

Plan: `syslog_bom_stripping_92c3c69f`. Not a ranked roadmap Open item; tracked as a maintenance/quality fix.

## Test plan

- [x] `rfc5424_strips_bom`, `rfc5424_bom_json_detected` (embedded-JSON regression), `rfc5424_keep_bom_when_disabled`, `bom_only_message` in the syslog adapter
- [x] `auto_detect_syslog_strips_bom` in the auto-detect adapter
- [x] `parse_input_format_threads_strip_bom` plus eval/daemon overlay tests (default-on, config-off, explicit-flag-wins)
- [x] `cargo test -p rsigma-runtime -p rsigma --all-features`
- [x] `mkdocs build --strict`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Each entry corresponds to a [GitHub Release](https://github.com/timescale/rsigma
 
 ## [Unreleased]
 
-### Strip the UTF-8 BOM from RFC 5424 syslog messages
+### Strip the UTF-8 BOM from RFC 5424 syslog messages (#187)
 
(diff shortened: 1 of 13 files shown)


sodomelle (approved):
It seems like the change is working as expected.

PR #191, opened 2026-06-09
feat: Fibratus conversion backend (rule YAML, fibratus_windows pipeline, sequence-DSL correlation)

### LINKED ISSUE #189: Add Fibratus backend to `rsigma-convert`
Add a `fibratus` target to `rsigma-convert` that converts Sigma rules into [Fibratus](https://github.com/rabbitstack/fibratus) rule YAML. Fibratus is an open-source Apache-2.0 Windows kernel-event detection and EDR engine; this would be rsigma's first conversion target aimed at endpoint sensors rather than log stores.

## Scope

- Detection rules emit Fibratus YAML (`name`, `id`, `description`, `labels`, `condition`, `min-engine-version`, optional `action`). A second `expr` format emits the bare filter expression.
- New builtin `fibratus_windows` pipeline maps Sigma Windows logsource categories to `evt.name` conditions (`process_creation` -> `CreateProcess`, `file_event` -> `CreateFile`, `image_load` -> `LoadModule`, `network_connection` -> `Connect`/`Send`, `registry_set` -> `RegSetValue`, ...) and renames PascalCase Sigma fields to the lowercase-dotted Fibratus vocabulary (`Image` -> `ps.exe`, `CommandLine` -> `ps.cmdline`, `TargetFilename` -> `file.path`, `TargetObject` -> `registry.path`, `DestinationIp` -> `net.dip`, ...).
- Recognized condition sub-trees are rewritten to upstream's [macro library](https://github.com/rabbitstack/fibratus/blob/master/rules/macros/macros.yml) (`spawn_process`, `open_file`, `modify_registry`, `create_remote_thread`, ...) when `-O use_macros=true` (default on). Matching runs on the parsed `ConditionExpr`, not on rendered strings.
- Sigma `temporal_ordered` correlation -> Fibratus `policy: sequence` with pattern bindings (`\$1.<field> = \$N.<field>`). Small-threshold `event_count`/`value_count` emulated via repeated/distinct slots (capped via `-O max_repeated_slots`). `value_sum`/`value_avg`/`value_percentile`/`value_median` and large thresholds raise structured `UnsupportedCorrelation` errors.
- Backend `-O` options: `action`, `min_engine`, `use_macros`, `default_logsource`, `emit_metadata`, `max_repeated_slots`, `temporal_permute`, `case_sensitive`.

## Notable challenges

- **Case-sensitivity inversion.** Sigma defaults to case-insensitive but Fibratus's `=`/`contains`/`startswith`/`endswith`/`matches`/`in`/`intersects` are case-sensitive; the `i`-prefixed variants are the insensitive ones. The backend populates `TextQueryConfig`'s default operator slots with the `i`-prefixed forms and the `case_sensitive_*_expression` slots with the bare forms (inverse of how other backends use those slots), plus overrides `convert_condition_as_in_expression` to pick `iin` vs `in` from item modifiers.
- **Regex is a function call, not an operator.** Sigma `|re` lowers to the [`regex(field, 'pat1', 'pat2', ...) = true`](https://fibratus.io/docs/rules/functions) filter function. Multi-value `|re` lists collapse into a single call; negated forms use a leading `not`. RE2 differs from PCRE on a few constructs (lookarounds, backreferences); patterns that use those return `UnsupportedModifier` rather than emit something Fibratus would reject at load time.
- **YAML envelope, not a query string.** `finalize_query` builds a per-rule YAML document and `finalize_output` joins with `---`; Phase 1 verifies whether the Fibratus loader accepts multi-doc YAML on stdin, with a CLI `directory` mode as the fallback.

Fibratus has a native `not` operator and no stream-selector/parser envelope, so this is smaller than the planned Loki backend (no De Morgan push-down, no stream selector) and requires no shared `ConversionState` changes.

## Out of scope

- A `directory` output mode that writes one `<tactic>_<rule>.yml` per rule. Defer to a follow-up once the multi-doc loader question is resolved.
- Reverse conversion (Fibratus YAML -> Sigma); tracked separately under the reverse-converter work.
- Linux/macOS Fibratus support (upstream is Windows-only).

mostafa:
## Summary

Adds a `fibratus` conversion target to `rsigma-convert` that lowers Sigma rules into rule YAML for [Fibratus](https://github.com/rabbitstack/fibratus), an Apache-2.0 kernel-event detection and EDR engine. Closes #189. This is the first conversion target aimed at an endpoint sensor rather than a centralized log store; the emitted rules drop directly into a Fibratus installation's `Rules/` directory and parse with the same loader as the upstream rules library. The mapping was reviewed by the Fibratus author (@rabbitstack) and reflects Fibratus 3.0.0 semantics.

## What is included

- **Backend** (`crates/rsigma-convert/src/backends/fibratus/`): `FibratusBackend`, `FibratusConfig`, the YAML envelope builder, the macro library + recognizer, and the Fibratus-flavored `TextQueryConfig`. The default operator slots carry the case-insensitive forms and the `case_sensitive_*` slots carry the bare forms, inverse to how other backends use those slots, because Sigma defaults to case-insensitive matching while Fibratus's plain operators are case-sensitive. Plain literal equality lowers to `field ~= 'literal'` (Fibratus's case-insensitive string-equality operator), or `field = 'literal'` with `|cased`; `imatches`/`matches` are reserved for values that actually carry `*`/`?` wildcards. The `evt.name` event discriminator always uses the exact `=`.
- **Builtin pipeline** (`crates/rsigma-eval/pipelines/fibratus_windows.yml`, registered next to `ecs_windows` and `sysmon`): maps the 14 supported Sigma Windows logsource categories to the right `evt.name` discriminators and renames Sigma's PascalCase fields to the lowercase-dotted Fibratus vocabulary. Fibratus 3.0.0 unified process attributes under `ps.*`, so `process_creation` maps the created child to `ps.exe` / `ps.cmdline` / `ps.pid` / `ps.username` and the spawner to `ps.parent.*`; `process_access` maps the opened process to the event arguments `evt.arg[exe]` / `evt.arg[pid]` (matching the upstream LSASS-access rule) and `GrantedAccess` to `ps.access.mask.names`.
- **Macro recognition** (`-O use_macros`, default `true`): walks top-level `and` clauses and replaces recognized runs with idiomatic Fibratus macro calls (`spawn_process`, `create_thread`, `write_file`, `read_file`, `open_file`, `create_file`, `set_value`, `open_process`, `open_thread`, ...). Greedy-longest-match so a full three-clause `open_file` triple beats the standalone `evt.name = 'CreateFile'` prefix; each clause is matched against both the exact (`=`) and case-insensitive (`~=`) forms, so recognition is independent of `-O case_sensitive`; unmatched clauses pass through verbatim.
- **Correlation lowering** (`backends/fibratus/correlation.rs`): `temporal_ordered` and `temporal` lower to Fibratus 1.10+'s inline `sequence ... maxspan ... by <fields> | stage | | stage |` DSL with one `|...|` stage per referenced rule. The shared `group-by` is emitted once as a sequence-level `by field1, field2, ...` clause (upstream rules-library style), so multi-field group-by needs no inline bindings. Small-threshold `event_count` and `value_count` expand into repeated/distinct stages capped at `-O max_repeated_slots` (default 5), with `value_count` distinctness expressed via positional pattern bindings (`$1`/`$2`). With `-O temporal_permute=true` and N <= 3 referenced rules, any-order `temporal` expands into one ordered sequence per permutation (1/2/6 documents). The four math-aggregate types, thresholds above the cap, range/equality predicates, and multi-rule `event_count` / `value_count` all return `UnsupportedCorrelation` with operator-actionable rationales.
- **Correlation window modes** (#192): honors the `rsigma.window` / `rsigma.gap` extension attributes. `sliding` (the default) is a faithful pass-through because the `sequence ... maxspan` DSL is itself a sliding total-span constraint per stage. `tumbling` returns `UnsupportedCorrelation` because Fibratus has no calendar-aligned bucket primitive. `session` emits a sliding sequence and pushes a warning that the requested per-step `rsigma.gap` is not enforced (no `maxpause`-style primitive); two new options pair with it: `-O correlation_method=sliding|session` (pySigma-style override, takes precedence over the rule's own `rsigma.window`; `tumbling` is rejected up-front) and `-O gap=<duration>` (default session gap for rules that do not declare their own).
- **ATT&CK labels**: `attack.t<NNNN>` Sigma tags lower to `technique.id` / `technique.ref` and `attack.t<NNNN>.<sub>` to `subtechnique.id` / `subtechnique.ref`, mirroring the upstream Fibratus convention. `attack.<tactic>` tags lower to `tactic.id` / `tactic.ref` / `tactic.name`.
- **Multi-value lists**: a multi-value OR list collapses into a single Fibratus list-operator clause (`field iin ('a', 'b')` for literals, `field imatches ('a*', 'b?')` when any value carries a wildcard, `field icontains (...)` / `istartswith` / `iendswith` for the substring modifiers). `|re` collapses into a single variadic `regex(field, p1, p2, ...)` call and `|cidr` into a single `cidr_contains(field, c1, c2, ...)` call. `|all` (AND, "all of") stays AND-joined because a list right-hand side is OR-only.
- **CLI integration** (`crates/rsigma-cli/src/commands/convert.rs`): `rsigma backend convert -t fibratus`, `rsigma backend targets`, the unknown-target error help text, and the `--target` flag help blurb all learn about `fibratus`. The CLI's per-rule output joining now defers to `backend.finalize_output(...)` instead of hard-coding `\n`, which fixes a latent bug for PostgreSQL's `view` / `continuous_aggregate` formats and is what makes the Fibratus `---` document separator land correctly on stdout.
- **Docs sweep**: a new [Fibratus backend reference](docs/reference/backends/fibratus.md) (modifier matrix, correlation matrix incl. window modes, options table, output formats with worked examples, caveats), nav entry, a Fibratus section in the [Rule Conversion guide](docs/guide/rule-conversion.md), an entry in the [Adding Backends](docs/developers/adding-backends.md) developer guide, root + convert-crate README updates, and a `[Unreleased]` CHANGELOG entry.

## Modifier coverage

| Sigma feature | Fibratus filter expression |
|---|---|
| Field equality (literal, no wildcards) | `field ~= 'value'` (case-insensitive default); `field = 'value'` with `\|cased`; `evt.name` always uses `=` |
| `contains` / `startswith` / `endswith` | `field icontains 'value'`, `field istartswith 'value'`, `field iendswith 'value'` (bare forms with `\|cased`) |
| Wildcards `*` / `?` | `field imatches '*pat?ern*'`; bare `matches` with `\|cased` |
| Multi-value list (OR / "any of") | `field iin ('a', 'b')`, or `field imatches ('a*', 'b?')` when any value has a wildcard, or `icontains`/`istartswith`/`iendswith` lists |
| Multi-value list with `\|all` ("all of") | AND-joined separate clauses (a list RHS is OR-only) |
| `re` modifier | `regex(field, 'pattern') = true`; multi-value collapses into a single variadic call. RE2 lookarounds and backreferences are rejected up-front with `UnsupportedModifier`. |
| `cidr` modifier | `cidr_contains(field, '10.0.0.0/8')`; multi-value collapses into a single variadic call |
| `gt` / `gte` / `lt` / `lte` | `field > N`, `field >= N`, `field < N`, `field <= N` |
| `exists` / `null` | `field != false` / `field = false`; a `null` value compares `field = ''` (Fibratus has no `null` token) |
| `fieldref` | `field1 = field2` (native) |
| AND/OR/NOT | Lowercase tokens; OR groups inside AND explicitly parenthesized so standard Sigma precedence survives |
| Keywords (unbound) | `UnsupportedKeyword` (Sigma keywords have no field; Fibratus operators require a bound one) |

## Correlation coverage

| Sigma type / attribute | Fibratus | Notes |
|---|---|---|
| `temporal_ordered` | `sequence` with one `\|...\|` stage per referenced rule and a single sequence-level `by <fields>` | First-class |
| `temporal` (any-order) | Same shape (ordered fallback documented in the description) | `-O temporal_permute=true` emits one sequence per permutation when N <= 3 |
| `event_count`, `gte`/`gt`, threshold <= cap | `sequence` with N repeated stages | Cap via `-O max_repeated_slots=5` |
| `value_count`, single field, threshold <= cap | `sequence` with N stages + positional pairwise inequality (`field != $1.field`) | Cap via `-O max_repeated_slots=5` |
| `value_sum` / `value_avg` / `value_percentile` / `value_median` | `UnsupportedCorrelation` | No Fibratus primitive |
| `rsigma.window: sliding` (default) | Native: `sequence ... maxspan <timespan>` | -- |
| `rsigma.window: tumbling` | `UnsupportedCorrelation` | No calendar-aligned bucket primitive |
| `rsigma.window: session` + `rsigma.gap` | Sliding sequence + warning | Gap not enforced (no `maxpause` primitive); `-O gap=<duration>` provides a default |

## Test plan

- `cargo test --workspace --all-features` -- all non-flaky suites pass, including:
  - 121 unit tests under `backends::fibratus::*` covering modifier mapping (`~=`/`=` literal equality, `evt.name = '...'`, wildcards to `imatches`/`matches`, the `|cased` switch), multi-value list collapsing (`iin`/`imatches`/`icontains` lists, single `cidr_contains` / `regex` calls, `|all` staying AND-joined), the YAML folded `condition: >` envelope, ATT&CK label mapping (techniques vs sub-techniques), the macro library + recognizer (greedy longest match, both `=`/`~=` forms, the `create_file` macro), correlation lowering with the sequence-level `by` clause and positional value-count bindings, `temporal_permute`, the three window modes (sliding pass-through, tumbling rejection, session degraded + warning) with operator-override coverage, and the unsupported-type rejections with structured errors.
  - 3 integration tests under `cli_convert` covering end-to-end conversion through the `fibratus_windows` pipeline + backend (`expr` format, `default` YAML envelope, and `-O action=kill` reaching the envelope), exercising the unified `ps.*` mappings.
- `cargo fmt --all -- --check` clean.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean (stable toolchain).
- `RUSTDOCFLAGS="-D warnings -D rustdoc::broken-intra-doc-links" cargo doc --workspace --all-features --locked --no-deps` clean.
- `mkdocs build --strict` clean.

CC @rabbitstack


crates/rsigma-cli/src/commands/convert.rs
```diff
@@ -213,6 +225,7 @@ pub(crate) fn cmd_list_targets() {
     println!("Available conversion targets:");
     println!("  postgres  - PostgreSQL/TimescaleDB (aliases: postgresql, pg)");
     println!("  lynxdb    - LynxDB log analytics engine");
+    println!("  fibratus  - Fibratus Windows kernel-event detection engine");
```
rabbitstack: Do we want to marry it to Windows? :D 
```suggestion
    println!("  fibratus  - Fibratus kernel-event detection engine");
```

crates/rsigma-cli/tests/cli_convert.rs
```diff
@@ -108,6 +108,102 @@ fn convert_with_format_timescaledb() {
...
+    );
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    // `evt.name: CreateProcess` injected by the `fibratus_windows`
+    // pipeline is recognized as the `spawn_process` macro under the
+    // default `-O use_macros=true`.
+    assert!(
+        stdout.contains("spawn_process"),
+        "stdout missing spawn_process: {stdout}",
+    );
+    // Sigma `process_creation.CommandLine` maps to
```
rabbitstack: Starting from Fibratus 3.0.0, `ps.sibling.*` fields have been decommissioned and unified under the `ps.*` namespace. So, `ps.exe` will yield the full executable path of the created process.

crates/rsigma-cli/tests/cli_convert.rs
```diff
@@ -108,6 +108,102 @@ fn convert_with_format_timescaledb() {
...
+    // `ps.sibling.cmdline` (the spawned process's cmdline) because
+    // on a Fibratus `CreateProcess` event the new process lives
+    // under the `ps.sibling.*` namespace.
+    assert!(
+        stdout.contains("ps.sibling.cmdline icontains 'whoami'"),
```
rabbitstack:
```suggestion
        stdout.contains("ps.cmdline icontains 'whoami'"),
```

crates/rsigma-cli/tests/cli_convert.rs
```diff
@@ -306,6 +402,7 @@ fn list_targets() {
     Available conversion targets:
       postgres  - PostgreSQL/TimescaleDB (aliases: postgresql, pg)
       lynxdb    - LynxDB log analytics engine
+      fibratus  - Fibratus Windows kernel-event detection engine
```
rabbitstack:
```suggestion
      fibratus  - Fibratus kernel-event detection engine
```

crates/rsigma-convert/src/backends/fibratus/correlation.rs
```diff
@@ -0,0 +1,1565 @@
+//! Sigma correlation rule → Fibratus sequence DSL lowering.
+//!
+//! Fibratus 1.10 decommissioned the old `policy: sequence` block in favor
+//! of an inline DSL inside `condition:`:
+//!
+//! ```text
+//! sequence
```
rabbitstack, line 7: Just a note. The `by` clause supports event correlation by multiple join fields ( e.g. `by ps.name, ps.sesionid`). If all steps in the sequence share the same join key(s), the `by` clause can be declared at the sequence level:

```
sequence
   maxspan 1m
   by ps.uuid
   ...
```

crates/rsigma-convert/src/backends/fibratus/correlation.rs
```diff
@@ -0,0 +1,1565 @@
...
+"#)
+        .unwrap();
+        assert_eq!(q.len(), 1);
+        let body = &q[0];
+        assert!(body.starts_with("sequence\nmaxspan 1m\n"));
+        // Per-stage bodies pass through the macro recognizer too, so
+        // `evt.name imatches 'Connect'` becomes `connect_socket`.
+        assert!(body.contains("|connect_socket\n"));
+        assert!(body.contains("| by ps.pid"));
+        assert!(body.contains("|spawn_process\n"));
```
rabbitstack: I'd adjust the test to ensure `spawn_process` has the trailing `by` clause

crates/rsigma-convert/src/backends/fibratus/correlation.rs
```diff
@@ -0,0 +1,1565 @@
...
+    - 00000000-0000-0000-0000-00000000000b
+  group-by:
+    - ps.pid
+    - ps.username
+  timespan: 30s
+"#)
+        .unwrap();
+        let body = &q[0];
+        // Primary field via the `by` clause, secondary via inline binding.
+        assert!(body.contains("| by ps.pid"));
```
rabbitstack: The resulting sequence should be a single top-level `by ps.pid, ps.username` with no bound fields.

crates/rsigma-convert/src/backends/fibratus/envelope.rs
```diff
@@ -0,0 +1,463 @@
...
+  selection:
+    ps.name: cmd.exe
+  condition: selection
+"#,
+        );
+        let out = render_rule_yaml(&r, "ps.name = 'cmd.exe'", &cfg());
+        let expected = "\
+name: Test Rule
+id: 12345678-1234-1234-1234-1234567890ab
+condition: ps.name = 'cmd.exe'
```
rabbitstack: Would it be possible to emit the condition as a folded block? For example:

```
condition: >
   ps.name = 'cmd.exe'
```

crates/rsigma-convert/src/backends/fibratus/macros.rs
```diff
@@ -0,0 +1,528 @@
...
+    ("duplicate_handle", "evt.name = 'DuplicateHandle'"),
+    ("create_handle", "evt.name = 'CreateHandle'"),
+    ("query_dns", "evt.name = 'QueryDns'"),
+    ("reply_dns", "evt.name = 'ReplyDns'"),
+    // Multi-clause macros, matched as contiguous clause runs.
+    (
+        "open_file",
+        "evt.name = 'CreateFile' and file.operation = 'OPEN' and file.status = 'Success'",
+    ),
+    (
```
rabbitstack, line 61: The `create_file` macro contemplates any creation disposition mask different than `OPEN`. It is the typical macro to express file creation conditions.

crates/rsigma-convert/src/backends/fibratus/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+        field1: &str,
+        field2: &str,
+        _state: &mut ConversionState,
+    ) -> Result<ConvertResult> {
+        let f1 = self.escape_and_quote_field(field1);
+        let f2 = self.escape_and_quote_field(field2);
+        Ok(ConvertResult::Query(format!("{f1} = {f2}")))
+    }
+
+    fn convert_keyword(&self, _value: &SigmaValue, _state: &mut ConversionState) -> Result<String> {
```
rabbitstack, line 738: If I correctly understood the Sigma keywords feature, it allows for performing a string search on multiple tokens. If this is right, in Fibratus, it can be expressed by enclosing the keywords inside parentheses. Examples:

```
ps.name in ('cmd.exe', 'pwsh.exe')

ps.exe imatches ('?:\\Windows\\System32\\*.exe', '?:\\Users\\*\\*.exe')
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+  a:
+    ps.name: cmd.exe
+  b:
+    ps.parent.name: explorer.exe
+  condition: a and b
+"#,
+    );
+    assert_eq!(
+        q,
+        vec!["ps.name imatches 'cmd.exe' and ps.parent.name imatches 'explorer.exe'"]
```
rabbitstack: If this detection doesn't involve glob matching, evaluation with string equality would be more efficient.

Case insensitive:

```
ps.name ~= 'cmd.exe' and ps.parent.name ~= 'explorer.exe'
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+  s:
+    ps.name:
+      - cmd.exe
+      - pwsh.exe
+  condition: s
+"#,
+    );
+    assert_eq!(
+        q,
+        vec!["(ps.name imatches 'cmd.exe' or ps.name imatches 'pwsh.exe')"]
```
rabbitstack: Multi-value lists can be evaluated with `in/iin` operators. For example:

```
ps.name iin ('cmd.exe', 'pwsh.exe')
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+  s:
+    ps.cmdline|contains|all:
+      - whoami
+      - localgroup
+  condition: s
+"#,
+    );
+    assert_eq!(
+        q,
+        vec!["ps.cmdline icontains 'whoami' and ps.cmdline icontains 'localgroup'"]
```
rabbitstack, line 354: Can be merged into a single condition:

```
ps.cmdline icontains ('whoami', 'localgroup')
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+    let q = convert(
+        r#"
+title: T
+detection:
+  s:
+    ps.username: null
+  condition: s
+"#,
+    );
+    assert_eq!(q, vec!["ps.username = null"]);
```
rabbitstack: `null` is an unrecognized token. We should compare against an empty string: 

```
ps.username = ''
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+      - '10.0.0.0/8'
+      - '172.16.0.0/12'
+      - '192.168.0.0/16'
+  condition: s
+"#,
+    );
+    assert_eq!(
+        q,
+        vec![
+            "(cidr_contains(net.dip, '10.0.0.0/8') or cidr_contains(net.dip, '172.16.0.0/12') or cidr_contains(net.dip, '192.168.0.0/16'))"
```
rabbitstack: `cidr_contains` accepts multiple CIDR masks, so a single condition would do it:

```
cidr_contains(net.dip, '10.0.0.0/8', '172.16.0.0/12', '192.168.0.0/16')
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+    let q = convert(
+        r#"
+title: T
+detection:
+  s:
+    thread.callstack.is_unbacked|exists: true
+  condition: s
+"#,
+    );
+    assert_eq!(q, vec!["thread.callstack.is_unbacked != null"]);
```
rabbitstack:
```suggestion
    assert_eq!(q, vec!["thread.callstack.is_unbacked != false"]);
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+    let q = convert(
+        r#"
+title: T
+detection:
+  s:
+    thread.callstack.is_unbacked|exists: false
+  condition: s
+"#,
+    );
+    assert_eq!(q, vec!["thread.callstack.is_unbacked = null"]);
```
rabbitstack:
```suggestion
    assert_eq!(q, vec!["thread.callstack.is_unbacked = false"]);
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+    let backend = FibratusBackend::new();
+    let rule = &mut collection.rules[0];
+    let state = apply_pipelines_with_state(&[pipeline], rule).unwrap();
+    let q = backend.convert_rule(rule, "expr", &state).unwrap();
+
+    assert_eq!(q.len(), 1);
+    let out = &q[0];
+    assert!(out.contains("spawn_process"), "got: {out}");
+    // On a Fibratus `CreateProcess` event the spawned process lives
+    // under `ps.sibling.*` (Sigma `Image` -> `ps.sibling.exe`,
```
rabbitstack: Not applicable starting from 3.0.0. Same comment I left at the very beginning.

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+    // Sigma `CommandLine` -> `ps.sibling.cmdline`) and the parent
+    // (event generator) is `ps.exe`. The Sigma source `'\cmd.exe'`
+    // parses as the literal `\cmd.exe`; Fibratus single-quoted
+    // strings need `\\` for a literal `\`, so the rendered value
+    // carries the double-escape.
+    assert!(
+        out.contains(r"ps.sibling.exe iendswith '\\cmd.exe'"),
```
rabbitstack:
```suggestion
        out.contains(r"ps.exe iendswith '\\cmd.exe'"),
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+        "got: {out}",
+    );
+    assert!(
+        out.contains(r"ps.exe iendswith '\\explorer.exe'"),
+        "got: {out}"
+    );
+    assert!(
+        out.contains("ps.sibling.cmdline icontains 'whoami'"),
```
rabbitstack:
```suggestion
        out.contains("ps.cmdline icontains 'whoami'"),
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+    // (event generator) is `ps.exe`. The Sigma source `'\cmd.exe'`
+    // parses as the literal `\cmd.exe`; Fibratus single-quoted
+    // strings need `\\` for a literal `\`, so the rendered value
+    // carries the double-escape.
+    assert!(
+        out.contains(r"ps.sibling.exe iendswith '\\cmd.exe'"),
+        "got: {out}",
+    );
+    assert!(
+        out.contains(r"ps.exe iendswith '\\explorer.exe'"),
```
rabbitstack:
```suggestion
        out.contains(r"ps.parent.exe iendswith '\\explorer.exe'"),
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+    let rule = &mut collection.rules[0];
+    let state = apply_pipelines_with_state(&[pipeline], rule).unwrap();
+    let q = backend.convert_rule(rule, "expr", &state).unwrap();
+
+    let out = &q[0];
+    // `evt.name: RegSetValue` is the single-clause `set_value` macro
+    // body but the registry status guard is not added by the
+    // pipeline, so the standalone clause stays as the raw form.
+    assert!(
+        out.contains("evt.name imatches 'RegSetValue'"),
```
rabbitstack: Should be a direct match:
```
evt.name = 'RegSetValue'
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+    let state = apply_pipelines_with_state(&[pipeline], rule).unwrap();
+    let q = backend.convert_rule(rule, "expr", &state).unwrap();
+    assert_eq!(q.len(), 1);
+    let out = &q[0];
+    assert!(
+        out.contains("spawn_process"),
+        "expected spawn_process, got: {out}"
+    );
+    assert!(
+        !out.contains("evt.name imatches 'CreateProcess'"),
```
rabbitstack:
```suggestion
        !out.contains("evt.name = 'CreateProcess'"),
```

crates/rsigma-convert/src/backends/fibratus/tests.rs
```diff
@@ -0,0 +1,972 @@
...
+    Image|endswith: '\cmd.exe'
+  condition: selection
+"#;
+    let mut collection = rsigma_parser::parse_sigma_yaml(yaml).unwrap();
+    let pipeline = resolve_builtin("fibratus_windows").unwrap().unwrap();
+    let rule = &mut collection.rules[0];
+    let state = apply_pipelines_with_state(&[pipeline], rule).unwrap();
+    let q = backend.convert_rule(rule, "expr", &state).unwrap();
+    assert!(
+        q[0].contains("evt.name imatches 'CreateProcess'"),
```
rabbitstack: Should be a direct match:

```suggestion
        q[0].contains("evt.name = 'CreateProcess'"),
```

crates/rsigma-convert/README.md
```diff
@@ -26,6 +26,7 @@ The crate provides a generic conversion framework that any backend can plug into
 | Test | `test` | Backend-neutral text queries for foundation testing |
 | PostgreSQL | `postgres`, `postgresql`, `pg` | Native PostgreSQL SQL with TimescaleDB support |
 | LynxDB | `lynxdb` | SPL2-compatible search queries for LynxDB log analytics engine |
+| Fibratus | `fibratus` | Fibratus rule YAML for the Apache-2.0 Windows kernel-event detection and EDR engine |
```
rabbitstack:
```suggestion
| Fibratus | `fibratus` | Fibratus rule YAML for the Apache-2.0 kernel-event detection and EDR engine |
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+# Usage:
+#   rsigma backend convert -t fibratus -p fibratus_windows rules/windows/process_creation/
+
+name: fibratus_windows
+priority: 20
+
+transformations:
+  # --- Process creation ---
+  #
+  # On a Fibratus `CreateProcess` event the *generating* process is
```
rabbitstack: As per previous comments, no longer applicable. Use the standard `ps.exe`, `ps.name`, etc. fields.

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+  # the parent that called the syscall; the new/child process lives
+  # under the `ps.sibling.*` namespace. That inverts the Sigma
+  # mapping every other category uses: Sigma `process_creation.Image`
+  # is the *spawned* (child) process, not the parent that spawned
+  # it.
+  - id: fibratus_process_creation_fields
+    type: field_name_mapping
+    mapping:
+      Image: ps.sibling.exe
```
rabbitstack: In Sigma, is `Image` an isolated process image name or a full executable path? Asking because `ps.name` returns the process executable base file name (e.g. `cmd.exe`)

rabbitstack:
```suggestion
      Image: ps.exe
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      CommandLine: ps.sibling.cmdline
```
rabbitstack:
```suggestion
      CommandLine: ps.cmdline
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      ProcessId: ps.sibling.pid
```
rabbitstack:
```suggestion
      ProcessId: ps.pid
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      User: ps.sibling.username
```
rabbitstack:
```suggestion
      User: ps.username
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      LogonId: ps.sibling.sessionid
```
rabbitstack:
```suggestion
      LogonId: ps.sessionid
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      ParentImage: ps.exe
```
rabbitstack:
```suggestion
      ParentImage: ps.parent.exe
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      ParentCommandLine: ps.cmdline
```
rabbitstack:
```suggestion
      ParentCommandLine: ps.parent.cmdline
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      ParentProcessId: ps.pid
```
rabbitstack:
```suggestion
      ParentProcessId: ps.parent.pid
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      User: ps.username
+    rule_conditions:
+      - type: logsource
+        category: file_event
+        product: windows
+
+  - id: fibratus_file_event_condition
+    type: add_condition
+    conditions:
+      evt.name: CreateFile
```
rabbitstack, line 117: Must map to `create_file` macro that ultimately expands to `evt.name = 'CreateFile' and file.operation != 'OPEN'`.

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      ProcessId: ps.pid
+      User: ps.username
+      DestinationIp: net.dip
+      DestinationHostname: net.host
+      DestinationPort: net.dport
+      DestinationPortName: net.dport_name
+      SourceIp: net.sip
+      SourcePort: net.sport
+      Protocol: net.l4_proto
+      Initiated: net.is_outbound
```
rabbitstack: This field doesn't exist.

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+    rule_conditions:
+      - type: logsource
+        category: network_connection
+        product: windows
+
+  # --- DNS query ---
+  - id: fibratus_dns_query_fields
+    type: field_name_mapping
+    mapping:
+      QueryName: net.dns.name
```
rabbitstack:
```suggestion
      QueryName: dns.name
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      QueryStatus: net.dns.rcode
```
rabbitstack:
```suggestion
      QueryStatus: dns.rcode
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      QueryResults: net.dns.answers
```
rabbitstack:
```suggestion
      QueryResults: dns.answers
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+    rule_conditions:
+      - type: logsource
+        category: dns_query
+        product: windows
+
+  # --- Image / DLL load ---
+  - id: fibratus_image_load_fields
+    type: field_name_mapping
+    mapping:
+      ImageLoaded: image.path
```
rabbitstack:
```suggestion
      ImageLoaded: module.path
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      OriginalFileName: image.path
```
rabbitstack:
```suggestion
      OriginalFileName: module.path
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      Image: ps.exe
+      ProcessId: ps.pid
+      Signed: image.signature.exists
```
rabbitstack:
```suggestion
      Signed: module.signature.exists
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      SignatureStatus: image.signature.status
```
rabbitstack: This field is not supported.

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      Signature: image.signature.subject
```
rabbitstack:
```suggestion
      Signature: module.signature.subject
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      Hashes: image.hashes
```
rabbitstack: Not supported

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      Imphash: image.imphash
```
rabbitstack: Not supported on image load events.

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+        category: image_load
+        product: windows
+
+  # --- Registry add/set/delete ---
+  - id: fibratus_registry_event_fields
+    type: field_name_mapping
+    mapping:
+      TargetObject: registry.path
+      Details: registry.value
+      EventType: registry.operation
```
rabbitstack: EvenType directly maps to Fibratus event type, such as `RegCreateKey` or `RegSetValue`

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      - type: logsource
+        category: image_load
+        product: windows
+
+  # --- Registry add/set/delete ---
+  - id: fibratus_registry_event_fields
+    type: field_name_mapping
+    mapping:
+      TargetObject: registry.path
+      Details: registry.value
```
rabbitstack:
```suggestion
      Details: registry.data
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+    rule_conditions:
+      - type: logsource
+        category: registry_event
+        product: windows
+
+  - id: fibratus_registry_set_fields
+    type: field_name_mapping
+    mapping:
+      TargetObject: registry.path
+      Details: registry.value
```
rabbitstack:
```suggestion
      Details: registry.data
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+    type: add_condition
+    conditions:
+      evt.name: RegDeleteKey
+    rule_conditions:
+      - type: logsource
+        category: registry_delete
+        product: windows
+
+  # --- Pipe created ---
+  - id: fibratus_pipe_created_fields
```
rabbitstack: Fibratus doesn't have visibility into pipe events.

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+    rule_conditions:
+      - type: logsource
+        category: create_remote_thread
+        product: windows
+
+  # --- Driver load ---
+  - id: fibratus_driver_load_fields
+    type: field_name_mapping
+    mapping:
+      ImageLoaded: image.path
```
rabbitstack:
```suggestion
      ImageLoaded: module.path
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      Image: image.path
```
rabbitstack:
```suggestion
      Image: module.path
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      Signed: image.signature.exists
```
rabbitstack:
```suggestion
      Signed: module.signature.exists
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      SignatureStatus: image.signature.status
```
rabbitstack: This field doens't exist

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      Signature: image.signature.subject
```
rabbitstack:
```suggestion
      Signature: module.signature.subject
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      Hashes: image.hashes
```
rabbitstack: This field doesn't exist

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+  # rule). `GrantedAccess` maps to `ps.access.mask.names`, the
+  # human-readable slice of access-right names that Sigma rules
+  # typically test against (`ps.access.mask` exposes the raw mask
+  # value if needed).
+  - id: fibratus_process_access_fields
+    type: field_name_mapping
+    mapping:
+      SourceImage: ps.exe
+      SourceProcessId: ps.pid
+      TargetImage: ps.sibling.exe
```
rabbitstack:
```suggestion
      TargetImage: evt.arg[exe]
```

crates/rsigma-eval/pipelines/fibratus_windows.yml
```diff
@@ -0,0 +1,389 @@
...
+      TargetProcessId: ps.sibling.pid
```
rabbitstack:
```suggestion
      TargetProcessId: evt.arg[pid]
```

docs/guide/rule-conversion.md
```diff
@@ -28,6 +29,7 @@ rsigma backend targets
 Available conversion targets:
   postgres  - PostgreSQL/TimescaleDB (aliases: postgresql, pg)
   lynxdb    - LynxDB log analytics engine
+  fibratus  - Fibratus Windows kernel-event detection engine
```
rabbitstack:
```suggestion
  fibratus  - Fibratus kernel-event detection engine
```

docs/guide/rule-conversion.md
```diff
@@ -320,6 +322,46 @@ rsigma backend convert rules/ -t lynxdb -p pipeline.yml
 # Output: FROM security_logs | search ...
 ```
 
+## Fibratus
+
+The Fibratus backend produces Fibratus rule YAML — the format consumed by [Fibratus](https://github.com/rabbitstack/fibratus), an Apache-2.0 Windows kernel-event detection and EDR engine. Unlike the other two backends, Fibratus is a runtime sensor rather than a log store, so the converted rules drop directly into a Fibratus installation's `Rules/` directory and the upstream loader runs them in-place.
```
rabbitstack:
```suggestion
The Fibratus backend produces Fibratus rule YAML — the format consumed by [Fibratus](https://github.com/rabbitstack/fibratus), an Apache-2.0 kernel-event detection and EDR engine. Unlike the other two backends, Fibratus is a runtime sensor rather than a log store, so the converted rules drop directly into a Fibratus installation's `Rules/` directory and the upstream loader runs them in-place.
```

docs/reference/backends/fibratus.md
```diff
@@ -0,0 +1,200 @@
+# Fibratus Backend
+
+The `fibratus` backend converts Sigma rules into [Fibratus](https://github.com/rabbitstack/fibratus) rule YAML, the rule format consumed by Fibratus's open-source Windows kernel-event detection and EDR engine. It is the first conversion target aimed at an endpoint sensor rather than a centralized log store; the produced rules drop directly into a Fibratus installation's `Rules/` directory and are accepted by the same loader that ships with the upstream rules library.
```
rabbitstack:
```suggestion
The `fibratus` backend converts Sigma rules into [Fibratus](https://github.com/rabbitstack/fibratus) rule YAML, the rule format consumed by Fibratus's open-source kernel-event detection and EDR engine. It is the first conversion target aimed at an endpoint sensor rather than a centralized log store; the produced rules drop directly into a Fibratus installation's `Rules/` directory and are accepted by the same loader that ships with the upstream rules library.
```

docs/reference/backends/fibratus.md
```diff
@@ -0,0 +1,200 @@
...
+
+```sh
+rsigma backend convert rules/windows/process_creation/ -t fibratus -p fibratus_windows
+```
+
+The pipeline maps logsource categories to `evt.name` discriminators and renames fields:
+
+| Sigma logsource | Fibratus `evt.name` | Representative field renames |
+|-----------------|---------------------|-------------------------------|
+| `process_creation` | `CreateProcess` | `Image -> ps.sibling.exe`, `CommandLine -> ps.sibling.cmdline`, `User -> ps.sibling.username` (Sigma's `process_creation.Image` is the *spawned* child process, which Fibratus exposes via the `ps.sibling.*` namespace on a `CreateProcess` event); `ParentImage -> ps.exe`, `ParentCommandLine -> ps.cmdline`, `ParentProcessId -> ps.pid` (the *generator* of a `CreateProcess` event is the parent) |
```
rabbitstack:
```suggestion
| `process_creation` | `CreateProcess` | `Image -> ps.exe`, `CommandLine -> ps.cmdline`, `User -> ps.username` (Sigma's `process_creation.Image` is the *spawned* child process, which Fibratus exposes via the `ps.*` namespace on a `CreateProcess` event); `ParentImage -> ps.parent.exe`, `ParentCommandLine -> ps.parent.cmdline`, `ParentProcessId -> ps.parent.pid` (the *generator* of a `CreateProcess` event is the parent) |
```

docs/reference/backends/fibratus.md
```diff
@@ -0,0 +1,200 @@
...
+| `process_termination` | `TerminateProcess` | `Image -> ps.exe`, `ProcessId -> ps.pid` |
+| `file_event` | `CreateFile` | `TargetFilename -> file.path`, `Image -> ps.exe` |
```
rabbitstack: Here, we should distinguish between file access and file creation events.

docs/reference/backends/fibratus.md
```diff
@@ -0,0 +1,200 @@
...
+| `file_delete` | `DeleteFile` | `TargetFilename -> file.path` |
+| `network_connection` | `Connect` | `DestinationIp -> net.dip`, `DestinationPort -> net.dport`, `SourceIp -> net.sip`, `Initiated -> net.is_outbound` |
+| `dns_query` | `QueryDns` | `QueryName -> net.dns.name`, `QueryStatus -> net.dns.rcode`, `QueryResults -> net.dns.answers` |
+| `image_load` | `LoadModule` | `ImageLoaded -> image.path`, `Signed -> image.signature.exists`, `Hashes -> image.hashes` |
+| `registry_set` | `RegSetValue` | `TargetObject -> registry.path`, `Details -> registry.value` |
+| `registry_add` | `RegCreateKey` | `TargetObject -> registry.path` |
+| `registry_delete` | `RegDeleteKey` | `TargetObject -> registry.path` |
+| `pipe_created` | `CreateFile` + `file.type = 'Pipe'` | `PipeName -> file.name` |
```
rabbitstack: Sadly, Fibratus can't tap into named pipe events. Hard to get any visibility without a kernel driver.

docs/reference/backends/fibratus.md
```diff
@@ -0,0 +1,200 @@
...
+| `create_remote_thread` | `CreateThread` | `SourceImage -> ps.exe`, `SourceProcessId -> ps.pid`, `TargetProcessId -> thread.pid`, `StartAddress -> thread.start_address`, `StartModule -> thread.start_address.module`, `StartFunction -> thread.start_address.symbol` (no `thread.image` field exists on `CreateThread` events; Sigma `TargetImage` rules fail conversion under this logsource) |
+| `driver_load` | `LoadModule` | `ImageLoaded -> image.path`, `Signed -> image.signature.exists` |
+| `process_access` | `OpenProcess` | `SourceImage -> ps.exe`, `SourceProcessId -> ps.pid`, `TargetImage -> ps.sibling.exe`, `TargetProcessId -> ps.sibling.pid`, `GrantedAccess -> ps.access.mask.names` (named slice; the upstream LSASS-dump rule pairs `open_process` with `ps.sibling.name ~= 'lsass.exe'`) |
```
rabbitstack:
```suggestion
| `process_access` | `OpenProcess` | `SourceImage -> ps.exe`, `SourceProcessId -> ps.pid`, `TargetImage -> evt.arg[exe]`, `TargetProcessId -> evt.arg[pid]`, `GrantedAccess -> ps.access.mask.names` (named slice; the upstream LSASS-dump rule pairs `open_process` with `ps.sibling.name ~= 'lsass.exe'`) |
```

docs/reference/backends/fibratus.md
```diff
@@ -0,0 +1,200 @@
...
+name: Suspicious cmd via Explorer
+id: 11111111-2222-3333-4444-555555555555
+description: |
+  Detect cmd.exe spawned by explorer.exe with whoami in args.
+labels:
+  tactic.id: TA0002
+  tactic.name: Execution
+  tactic.ref: 'https://attack.mitre.org/tactics/TA0002/'
+condition: >
+  ps.exe iendswith '\\cmd.exe' and ps.parent.exe iendswith '\\explorer.exe'
```
rabbitstack: Is the rule emitted with the event predicate being the last condition? If so, we should make it the very first condition in the rule, so the engine can short-circuit and avoid wasting evaluation cycles.

docs/reference/backends/fibratus.md
```diff
@@ -0,0 +1,200 @@
...
+labels:
+  tactic.id: TA0006
+  tactic.name: Credential Access
+  tactic.ref: 'https://attack.mitre.org/tactics/TA0006/'
+  technique.id: T1110.001
+  technique.ref: 'https://attack.mitre.org/techniques/T1110/001/'
+condition: >
+  sequence
+  maxspan 5m
+    |evt.name imatches 'AuthFail'
```
rabbitstack: Mind `by net.sip` can be set at the sequence level instead of each stage.

CHANGELOG.md
```diff
@@ -4,6 +4,24 @@ All notable changes to RSigma are documented in this file. Each entry correspond
 
 ## [Unreleased]
 
+### Fibratus conversion backend (#191)
+
+Convert Sigma rules into rule YAML for [Fibratus](https://github.com/rabbitstack/fibratus), an Apache-2.0 Windows kernel-event detection and EDR engine. Fibratus is the first conversion target aimed at an endpoint sensor rather than a centralized log store; rules emitted by `rsigma backend convert -t fibratus` drop into a Fibratus installation's `Rules/` directory and load with the same parser as the upstream rules library.
```
rabbitstack:
```suggestion
Convert Sigma rules into rule YAML for [Fibratus](https://github.com/rabbitstack/fibratus), an Apache-2.0 kernel-event detection and EDR engine. Fibratus is the first conversion target aimed at an endpoint sensor rather than a centralized log store; rules emitted by `rsigma backend convert -t fibratus` drop into a Fibratus installation's `Rules/` directory and load with the same parser as the upstream rules library.
```

CHANGELOG.md
```diff
@@ -4,6 +4,24 @@ All notable changes to RSigma are documented in this file. Each entry correspond
...
...stry.path`, `DestinationIp -> net.dip`, `ImageLoaded -> image.path`, `QueryName -> net.dns.name`. Two categories invert the parent/child mapping because Fibratus exposes the *target* process under the `ps.sibling.*` namespace: `process_creation` (the spawned child) and `process_access` (the process being opened) both route `Image`/`CommandLine`/`User` -> `ps.sibling.exe`/`cmdline`/`username`, with `ParentImage`/`ParentCommandLine`/`ParentProcessId` -> `ps.exe`/`cmdline`/`pid` (the event generator). `process_access` also maps `GrantedAccess -> ps.access.mask.names` (the named slice of access rights). Use it whenever you convert SigmaHQ Windows rules: `rsigma backend convert rules/windows/ -t fibratus -p fibratus_windows`. ATT&CK tags in `tags:` flatten into Fibratus's `labels:` block via a static MITRE lookup: `attack.<tactic_short_name>` -> `tactic.id`/`tactic.name`/`tactic.ref`, `attack.t<NNNN>` -> `technique.id`/`technique.ref`, and `attack.t<NNNN>.<sub>` -> `subtechnique.id`/`subtechnique.ref` (the base technique and sub-technique live in separate label namespaces matching the upstream Fibratus rules library convention). Unknown tags pass through as `tag.<original>: <original>`.
```
rabbitstack:
```suggestion
**Field naming.** A new `fibratus_windows` builtin pipeline (registered alongside `ecs_windows` and `sysmon`) maps Sigma's PascalCase Windows fields to the lowercase-dotted Fibratus vocabulary and adds the right `evt.name` discriminator per logsource category (`process_creation -> CreateProcess`, `network_connection -> Connect`, `dns_query -> QueryDns`, `registry_set -> RegSetValue`, ...). Most categories map `Image -> ps.exe`, `CommandLine -> ps.cmdline`, `TargetFilename -> file.path`, `TargetObject -> registry.path`, `DestinationIp -> net.dip`, `ImageLoaded -> image.path`, `QueryName -> dns.name`. `process_access` also maps `GrantedAccess -> ps.access.mask.names` (the named slice of access rights). Use it whenever you convert SigmaHQ Windows rules: `rsigma backend convert rules/windows/ -t fibratus -p fibratus_windows`. ATT&CK tags in `tags:` flatten into Fibratus's `labels:` block via a static MITRE lookup: `attack.<tactic_short_name>` -> `tactic.id`/`tactic.name`/`tactic.ref`, `attack.t<NNNN>` -> `technique.id`/`technique.ref`, and `attack.t<NNNN>.<sub>` -> `subtechnique.id`/`subtechnique.ref` (the base technique and sub-technique live in separate label namespaces matching the upstream Fibratus rules library convention). Unknown tags pass through as `tag.<original>: <original>`.
```

mostafa:
Thanks for the thorough review, @rabbitstack. This is hugely valuable since it comes straight from the engine author. I pushed 70903d26 addressing all of it. Walking through each thread:

## Field namespace (3.0.0 `ps.*` unification)

Dropped the legacy `ps.sibling.*` mappings everywhere now that 3.0.0 unifies process attributes under `ps.*`:

- `process_creation`: `Image` -> `ps.exe`, `CommandLine` -> `ps.cmdline`, `ProcessId` -> `ps.pid`, `User` -> `ps.username` (the created process), and the spawner is `ParentImage` -> `ps.parent.exe`, `ParentCommandLine` -> `ps.parent.cmdline`, `ParentProcessId` -> `ps.parent.pid`.
- `process_access`: I checked the 3.0.0 `credential_access_lsass_access_from_unsigned_executable.yml` rule and saw the target is `evt.arg[exe]`, so `TargetImage` -> `evt.arg[exe]` and `TargetProcessId` -> `evt.arg[pid]`. `GrantedAccess` stays `ps.access.mask.names`.
- `Image: ps.exe`, `CommandLine: ps.cmdline`, `ProcessId: ps.pid` suggestions applied; test assertions and the doc field table updated to match.

## Operators

- String equality is no longer a glob. Literal values without wildcards now use `~=` (case-insensitive default) and `=` (with `|cased`), per your `ps.name ~= 'cmd.exe'` note. `imatches`/`matches` are reserved for values that actually carry `*`/`?`.
- `evt.name` always uses the exact `=` (your `evt.name = 'RegSetValue'` and `evt.name = 'CreateProcess'` notes), so it matches the macro library and upstream rules style.
- `null`: a Sigma `field: null` now lowers to `field = ''` since `null` is not a recognized token.
- `exists`: lowers to `field != false` / `field = false`.
- The callstack suggestions (`thread.callstack.is_unbacked != false` / `= false`) now fall out of that `exists` mapping.

## Multi-value lists

These now collapse into a single Fibratus list-operator clause instead of OR'd comparisons:

- `ps.name iin ('cmd.exe', 'pwsh.exe')` for literal equality.
- `ps.exe imatches ('?:\Windows\System32\*.exe', ...)` when any value carries a wildcard.
- `cidr_contains(net.dip, '10.0.0.0/8', '172.16.0.0/12', '192.168.0.0/16')` as a single variadic call.
- The `ps.cmdline icontains ('whoami', 'localgroup')` suggestion is applied for the OR case. One nuance: the specific test you flagged uses Sigma's `|all` modifier, which is AND ("contains all of"), and a list right-hand side is OR ("any of"), so collapsing that one would flip the semantics. Those `|all` values stay AND-joined as separate clauses; the plain (OR) multi-value lists do collapse.

## `by` clause in sequences

Reworked correlation lowering to emit a single sequence-level `by field1, field2` clause after `maxspan` (your `sequence / maxspan 1m / by ps.uuid` example), instead of per-stage `by` plus inline `$1.field = field` bindings. Multi-field group-by is now `by ps.pid, ps.username` with no bound fields, and `value_count` distinctness uses positional pattern bindings (`$1`/`$2`).

## Macros

Added the upstream `create_file` macro (`evt.name = 'CreateFile' and file.operation != 'OPEN' and file.status = 'Success'`) as the canonical file-creation macro, alongside `open_file`/`create_new_file`/`create_file_supersede`. The recognizer now matches each clause against both the exact (`=`) and case-insensitive (`~=`) operator forms.

## Folded condition block

`condition:` is now always emitted as a folded `>` block (`condition: >`) including short single-line conditions, matching the upstream rules-library style.

## Keywords

I kept `UnsupportedKeyword` here. Sigma keywords are unbound (no field, full-text across the whole event), and the Fibratus list forms you showed (`ps.name in (...)`, `ps.exe imatches (...)`) all require a bound field, so there is no faithful field-less lowering. A user who wants this can bind the search to a field via a pipeline. Happy to revisit if there is an idiomatic Fibratus way to express a field-less token search.

## Docs and CLI

- `fibratus  - Fibratus kernel-event detection engine` (dropped "Windows") in the CLI and the targets list.
- The convert-crate README table now reads "Fibratus rule YAML for the Apache-2.0 kernel-event detection and EDR engine".
- The backend reference, conversion guide, and CHANGELOG are updated for all of the above.

All gates pass (fmt, clippy, workspace tests, rustdoc, mkdocs strict). Please take another look when you have a chance.

crates/rsigma-eval/pipelines/fibratus_windows.yml
rabbitstack: @mostafa DNS fields still reference invalid field names

crates/rsigma-eval/pipelines/fibratus_windows.yml
rabbitstack: @mostafa `image.` is deprecated. We should migrate to `module.*`

crates/rsigma-eval/pipelines/fibratus_windows.yml
rabbitstack: @mostafa Unsupported field should be removed

crates/rsigma-eval/pipelines/fibratus_windows.yml
rabbitstack: @mostafa The field is not supported. Let's axe it.

crates/rsigma-eval/pipelines/fibratus_windows.yml
rabbitstack: @mostafa Same here.

README.md
```diff
@@ -32,7 +32,7 @@ For rule quality and editor integration, a built-in linter validates rules again
 * **Processing pipelines:** Use pySigma-compatible processing pipelines for field mapping, transformations, conditions, and finalizers
 * **Dynamic pipelines:** Populate any pipeline value from external sources (HTTP, files, commands, NATS) with template expansion, auto-refresh, and data extraction via jq, JSONPath, or CEL
 * **Post-evaluation enrichment:** Inject contextual data (asset info, IP reputation, identity, GeoIP, runbook URLs, ...) into detection and correlation results via four primitives (`template`, `lookup`, `http`, `command`) with kind-aware template namespaces, response cache, scope filtering, and hot-reload
-* **Rule conversion:** Convert rules into backend-native query strings via a pluggable backend trait (PostgreSQL/TimescaleDB SQL, LynxDB)
+* **Rule conversion:** Convert rules into backend-native query strings via a pluggable backend trait (PostgreSQL/TimescaleDB SQL, LynxDB SPL2, Fibratus rule YAML for Windows EDR sensors)
```
rabbitstack:
```suggestion
* **Rule conversion:** Convert rules into backend-native query strings via a pluggable backend trait (PostgreSQL/TimescaleDB SQL, LynxDB SPL2, Fibratus rule YAML for EDR sensors)
```

rabbitstack (approved):
LGTM after applying field name suggestions in `fibratus_windows.yml` pipeline.

mostafa:
Thanks for the follow-up pass, @rabbitstack. Pushed in the latest commit. I validated every pipeline field against the Fibratus 3.0.0 field registry (`pkg/filter/fields/fields_windows.go`) so the whole pipeline is load-valid now:

## DNS fields
`net.dns.*` was wrong. Fixed to `QueryName -> dns.name`, `QueryStatus -> dns.rcode`, `QueryResults -> dns.answers`.

## image_load / driver_load (`image.*` deprecated)
Migrated to `module.*`: `ImageLoaded -> module.path`, `Signed -> module.signature.exists`, `Signature -> module.signature.subject`. `OriginalFileName -> pe.file.name`. The three you flagged (`SignatureStatus`, `Hashes`, `Imphash`) have no `module.*` equivalent in the registry, so I dropped them; a rule that references them now fails conversion rather than emitting an invalid field. Applied the same migration to `driver_load` for consistency. (Note: `pe.imphash` does exist if you'd ever want `Imphash` remapped there instead of dropped, but I followed your "axe it" for now.)

## Event predicate first
Good call on short-circuiting. I added a `prepend` option to the `add_condition` pipeline transformation and set it on every `fibratus_windows` event-name injection, so the rule now leads with the discriminator: `spawn_process and ps.exe iendswith '\cmd.exe' and ...` instead of trailing it. This is scoped to the Fibratus pipeline and does not change the other backends' output.

## by clause at sequence level
Already done in the previous commit (single top-level `by net.sip` rather than per-stage), and I updated the docs example to match.

## CHANGELOG / README
Applied your suggestions: the CHANGELOG first line and field-naming paragraph (now `dns.name`/`module.path`), and the README feature bullet (`Fibratus rule YAML for EDR sensors`, dropped "Windows").

## Beyond what was flagged
While validating against the registry I also found and fixed invalid fields in two sections you didn't flag, since they'd fail at load the same way:
- `network_connection`: `net.dport_name -> net.dport.name`, `net.l4_proto -> net.l4.proto`; dropped `DestinationHostname` (no scalar `net.*` field) and `Initiated` (no `net.is_outbound`).
- `registry_event`: dropped `EventType -> registry.operation` (no such field).

Flagging these explicitly in case you'd prefer a different mapping for any of them (e.g. `DestinationHostname -> net.dip.names`, which is a slice).

All gates pass (fmt, clippy, workspace tests, rustdoc, mkdocs strict), plus new pipeline tests for the `dns.*`/`module.*` mappings and the event-predicate-first ordering.

mostafa:
You're right, I missed a few on the previous pass. The latest commit covers the rest:

- **`file_event` -> `create_file`.** It now maps to file *creation*, not access: the `CreateFile` event plus a negated `OPEN` disposition guard (`evt.name = 'CreateFile' and ... and not (file.operation ~= 'OPEN')`), which is the `create_file` macro semantics. The pipeline's `add_condition` can only inject equality, so the guard is the negated `OPEN` form rather than the literal `file.operation != 'OPEN'`; the evaluated result is the same, though it doesn't render as the `create_file` macro name. If you'd rather I add operator-aware condition injection so it emits the macro verbatim, I'm happy to follow up.
- **`OriginalFileName -> module.path`.** Applied your suggestion (I had used `pe.file.name`).
- **`Details -> registry.data`.** Fixed in both `registry_set` and `registry_event` (was `registry.value`).
- **`pipe_created` removed.** Since Fibratus has no named-pipe visibility without a kernel driver, the logsource is now unmapped (rules fall through and fail conversion rather than aliasing to an unrelated event), and I removed it from the docs table with a caveat.
- **`docs/reference/backends/fibratus.md` intro.** Dropped "Windows" (`open-source kernel-event detection and EDR engine`).
- **`registry_event` EventType.** I removed the invalid `registry.operation` mapping. Mapping the `EventType` *value* to the corresponding `evt.name` (RegSetValue/RegCreateKey/...) needs value-based dispatch the generic transform can't express today, so the discriminator comes from the specific `registry_set`/`registry_add`/`registry_delete` categories instead; flagging in case you'd like a value-mapping follow-up for the generic `registry_event` category.

All gates pass (fmt, clippy, workspace tests, mkdocs strict), with new pipeline tests for the file_event OPEN-exclusion, plus the earlier dns.*/module.* and event-predicate-first coverage.

PR #201, opened 2026-06-12
feat(rstix): add leaf serde and common property model

SecurityEnthusiast:
## Summary

Introduces the first Phase 2 (`rstix`) data-model increment: shared common-property containers and leaf-type JSON serde, building on Phase 1 core primitives.

**Added**
- `rstix::model::common`: `SdoSroCommonProps`, `ScoCommonProps`, `ExternalReference`, `GranularMarking`, `ExtensionType` / `ExtensionEntry` / `ExtensionMap`
- `ModelError` for construction/deserialize invariants (`source_name` non-empty; granular marking `marking_ref` XOR `lang`)
- Leaf JSON serde: `StixId` + 42 typed IDs (prefix check on deserialize), `StixTimestamp`, `TaxiiTimestamp`, `LanguageTag`
- Integration tests: `crates/rstix/tests/spec.rs` with fixtures under `tests/fixtures/spec/common/`

**Serde layout**
- Hand-written: `serde_impls/` (`StixId`, timestamps)
- Macro-generated: typed ID wrappers in `define_typed_id!`
- Inline (like `SpecVersion`): `LanguageTag`

**Unchanged**
- Phase 1 `parse_bundle()` still returns `ParseError::NotImplemented` until `Bundle` lands
- No concrete SDO/SCO/SRO/Meta object types yet

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --workspace --all-features --locked` (1975 tests locally)
- [x] `mkdocs build --strict`
- [ ] CI green on this PR

## Docs

- `crates/rstix/README.md`, `docs/library/rstix.md`, `docs/reference/architecture.md`
- Root `CHANGELOG.md` `[Unreleased]` entry


mostafa (commented):
Reviewed at dae6c6f. I also ran the gates locally: `cargo fmt --check`, `cargo clippy -p rstix --all-targets --all-features -- -D warnings`, `cargo check -p rstix --no-default-features`, `cargo test -p rstix --all-features` (55 tests), and `mkdocs build --strict` all pass.

Nice slice overall: the `serde` feature gating holds up under `--no-default-features`, keeping typed-ID serde single-sourced in `define_typed_id!` is the right call, and the negative-path fixtures (missing/empty/whitespace `source_name`, both/neither granular marking, missing `spec_version`) are exactly the right coverage.

Inline: two spec-conformance items (confidence range, required `selectors`), one decision question on the external-reference at-least-one rule, and two minor non-blocking notes.

crates/rstix/src/model/common/sdo_sro.rs
```diff
@@ -0,0 +1,112 @@
...
+        feature = "serde",
+        serde(default, skip_serializing_if = "Vec::is_empty")
+    )]
+    pub labels: Vec<String>,
+    /// Confidence in the correctness of the data, `0..=100`.
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "Option::is_none")
+    )]
+    pub confidence: Option<u8>,
```
mostafa: Phase 1 already ships `core::Confidence` with a range-validated constructor (`ConfidenceError::OutOfRange` for values over 100). With a raw `u8` here, `"confidence": 255` deserializes successfully even though STIX requires `0..=100`, and the Phase 1 type goes unused in the model.

```suggestion
    /// Confidence in the correctness of the data, `0..=100`.
    #[cfg_attr(
        feature = "serde",
        serde(default, skip_serializing_if = "Option::is_none")
    )]
    pub confidence: Option<Confidence>,
```

This needs `Confidence` added to the `crate::core` import above, plus serde impls on `Confidence` (serialize the inner `u8`, deserialize via `Confidence::new` mapped through `de::Error::custom`). That fits this slice's leaf-type serde theme.

crates/rstix/src/model/common/granular.rs
```diff
@@ -0,0 +1,112 @@
...
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for GranularMarking {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        #[derive(serde::Deserialize)]
+        struct Raw {
+            #[serde(default)]
+            selectors: Vec<String>,
```
mostafa: `selectors` is required per STIX 2.1 §7.2.3, but `#[serde(default)]` lets a granular marking with no `selectors` field parse into an empty vec, and `validate()` does not catch it.

```suggestion
            selectors: Vec<String>,
```

Consider also rejecting empty `selectors` in `validate()` so an explicit `"selectors": []` is caught too (STIX prohibits empty lists).

crates/rstix/src/model/common/external_ref.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub fn new(source_name: impl Into<String>) -> Result<Self, ModelError> {
+        let reference = Self {
+            source_name: source_name.into(),
+            ..Self::default()
+        };
+        reference.validate()?;
+        Ok(reference)
+    }
+
+    /// Validate the `source_name` invariant.
```
mostafa: STIX 2.1 §2.5.2: "In addition to the source_name property, at least one of the description, url, or external_id properties MUST be present." `validate()` only enforces `source_name`, and the `external-reference-minimal.json` fixture locks in acceptance of a source-name-only reference as tested behavior.

Is the leniency deliberate? Enforcing it would also break `new("capec")`, which constructs exactly such a reference, so it would need a new `ModelError` variant plus a constructor rework. If it is deliberate (deferred to the validation-pipeline slice), documenting the decision would prevent it from reading as an oversight:

```suggestion
    /// Validate the `source_name` invariant.
    ///
    /// STIX 2.1 §2.5.2 additionally requires at least one of `description`,
    /// `url`, or `external_id` alongside `source_name`. That rule is not
    /// enforced here; it is deferred to the validation-pipeline slice.
```

crates/rstix/src/model/common/extension.rs
```diff
@@ -0,0 +1,122 @@
...
+/// use rstix::model::common::{ExtensionEntry, ExtensionMap, ExtensionType};
+///
+/// let mut map = ExtensionMap::default();
+/// map.0.insert(
+///     "extension-definition--00000000-0000-0000-0000-000000000001".into(),
+///     ExtensionEntry {
+///         extension_type: Some(ExtensionType::PropertyExtension),
+///         properties: Default::default(),
+///     },
+/// );
```
mostafa, line 99: Minor, non-blocking: the doc example reaching through `.0` reads awkwardly for the main public entry point. An `insert` method on `ExtensionMap` would read better and leaves room to add invariants later:

```suggestion
/// let mut map = ExtensionMap::default();
/// map.insert(
///     "extension-definition--00000000-0000-0000-0000-000000000001",
///     ExtensionEntry {
///         extension_type: Some(ExtensionType::PropertyExtension),
///         properties: Default::default(),
///     },
/// );
```

with something like `pub fn insert(&mut self, id: impl Into<String>, entry: ExtensionEntry) -> Option<ExtensionEntry> { self.0.insert(id.into(), entry) }` in the impl block below.

crates/rstix/tests/support/roundtrip.rs
```diff
@@ -0,0 +1,31 @@
...
+/// return the parsed value.
+pub fn roundtrip<T>(relative_path: &str) -> T
+where
+    T: DeserializeOwned + Serialize + PartialEq + Debug,
+{
+    let json = load_spec_fixture(relative_path);
+    let parsed: T = serde_json::from_str(&json).expect("deserialize");
+    let reserialized = serde_json::to_string(&parsed).expect("serialize");
+    let reparsed: T = serde_json::from_str(&reserialized).expect("reparse");
+    assert_eq!(parsed, reparsed);
```
mostafa, line 22: Non-blocking note: this helper cannot detect dropped fields, since it compares `parsed` against `reparsed` but never the re-serialized JSON against the original fixture. Today that is intentional (fixtures carry `type`/`name`/`value` that the common-prop structs deliberately ignore), but once concrete objects land in later slices, comparing `serde_json::Value`s against the original fixture would catch lossiness for free.

CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ The correlation engine's memory bounds are now fully operator-configurable. Prev
 - **API.** `CorrelationConfig` gains `max_group_entries: Option<usize>`, `CompiledCorrelation` gains the matching per-rule override, and `WindowState` gains `truncate_oldest(cap, preserve_front)`. The per-rule attribute follows the same resolution order as the other `rsigma.*` correlation attributes: rule override wins over the engine default.
 - **Docs.** New flags documented on the `engine eval` and `engine daemon` CLI pages, the configuration file reference, the custom-attributes reference, the processing-pipelines attribute table, and the Performance Tuning guide's correlation-memory section (which also now states that the global cap bounds group count, not bytes within a group, and is global rather than per-rule).
 
+### `rstix`: Phase 2 slice 1 — model skeleton and common properties
```
mostafa: Move this to the top.

crates/rstix/src/model/common/sdo_sro.rs
SecurityEnthusiast: Done — Option<Confidence> + serde_impls/confidence.rs.

crates/rstix/src/model/common/granular.rs
SecurityEnthusiast: Required selectors on deserialize, empty rejected in validate(), with_* delegates to new()

crates/rstix/src/model/common/external_ref.rs
SecurityEnthusiast: ExternalReferenceMissingDetail, updated minimal fixture, new(source, desc, url, ext_id)

crates/rstix/tests/support/roundtrip.rs
SecurityEnthusiast: to_value compared against fixture; subset compare for common-prop fixtures with extra SDO keys.

mostafa: Verified, and the fixture comparison is a nice upgrade. One doc nit, fine to fix in a later slice: the comment says standalone fixtures (for example ExternalReference) use full value equality, but the (Object, Object) arm applies the subset rule to every object fixture, so a dropped field would still go unnoticed for those too. Worth rewording so the comment does not overpromise.

mostafa (approved):
Re-reviewed at 167a15cf. All six review threads are addressed and resolved:

- confidence is now Option<Confidence> with range-validating serde in serde_impls/confidence.rs, with in-range and out-of-range fixtures
- selectors is required on deserialize and validate() rejects empty lists (GranularMarkingEmptySelectors), with both negative fixtures; with_marking_ref/with_lang delegating to new() keeps the invariants in one place
- the §2.5.2 at-least-one rule is enforced via ExternalReferenceMissingDetail with a reworked constructor, updated minimal fixture, and a source-only reject fixture
- ExtensionMap::insert added and the doc example no longer reaches through .0
- the roundtrip helper now compares re-serialized output against the fixture (one doc-wording nit left on the thread, fine for a later slice)
- CHANGELOG entry moved to the top of Unreleased

Verified locally on this head: cargo fmt --check, cargo check -p rstix --no-default-features, cargo clippy -p rstix --all-targets --all-features -- -D warnings, cargo test -p rstix --all-features (59 tests), and mkdocs build --strict all pass.

SecurityEnthusiast:
Thanks @mostafa

PR #202, opened 2026-06-12
fix(convert,fibratus): Correct Fibratus field mappings

rabbitstack:
## Summary

Fixes several issues in the Fibratus rule converter related to missing field mappings and improper event scoping.

## Changes

- Add missing process field mappings in the Fibratus converter
- Add missing `TargetImage` field mapping in thread events
- Scope registry events by event type name or category, as Fibratus rejects rules that omit event type scoping

## Notes

Without explicit event type scoping (by name or category), Fibratus will reject the converted rules at load time. This fix ensures registry events are always emitted with the required scope.


mostafa (approved):
@rabbitstack Thank you for your contribution! LGTM. 🙏

PR #204, opened 2026-06-13
feat(proto): portable Sigma IR protobuf interchange + conformance vectors

mostafa:
## Why

Sigma backends are reimplemented per engine today: a Splunk/KQL/Elastic/etc. backend written for one engine has to be re-ported into RSigma and any future engine. This defines a shared, strongly typed message so a backend can be written once and reused across engines, including as a remote gRPC service. The message sits at the post-pipeline, modifier-resolved, selector-resolved layer that pySigma (after modifier application and condition postprocess) and RSigma (its HIR) independently arrive at, which is what makes a single schema faithful rather than forced. External/dynamic sources are converging across engines too (see [SigmaHQ/pySigma#470](https://github.com/SigmaHQ/pySigma/issues/470)), so they belong in the shared schema rather than as an engine-specific extension.

## Summary

- Adds `proto/sigma_ir.proto`, a language-neutral protobuf schema for a post-pipeline, modifier-resolved, selector-resolved Sigma rule: the wire message a remote backend responds to, and a single canonical form an engine's HIR can be generated from or conformance-locked against.
- Adds `proto/sigma_backend.proto`, a `SigmaBackend` gRPC service (`Capabilities` + `Convert`, with explicit `Unsupported` results) as the separable remote-backend transport over the schema.
- Adds 19 golden `(rule YAML -> canonical IR)` conformance vectors under `proto/conformance/vectors/`, plus READMEs documenting the schema, conventions, and vector format.
- Schema and docs only; no crate code changes. Built by reconciling the RSigma parser AST/HIR with pySigma's resolved types, conditions, modifiers, correlations, and filters, so both engines lower to and consume the same form.
- Records the intent to extract `proto/` into a neutral standalone schema repo so a second implementation can vendor it.

## Test plan

- [x] `protoc` (libprotoc 34.1) compiles `sigma_ir.proto` and `sigma_backend.proto` cleanly.
- [x] All 19 conformance vectors validate strictly against the schema (generated Python bindings + `json_format.ParseDict`, which rejects unknown fields).
- [x] `base64offset` / `windash` expansions in the vectors were computed with pySigma's exact algorithm, not hand-written.
- [ ] Reviewer: confirm the schema home (extract to a neutral repo vs keep in-tree).

### DIFF
--- a/proto/README.md
+++ b/proto/README.md
@@ -1,0 +1,49 @@
+# Sigma portable IR (protobuf interchange)
+
+A language-neutral schema for a **post-pipeline, modifier-resolved, selector-resolved** Sigma rule. It is the single source of truth that an engine's internal HIR is generated from or conformance-locked against, and the wire message a remote backend (pySigma, RSigma, or any future engine) responds to.
+
+The schema reconciles the RSigma parser AST/HIR with pySigma's resolved types, conditions, modifiers, correlations, and filters, so both engines lower to and consume the same form.
+
+## Files
+
+| File | Purpose |
+|------|---------|
+| `sigma_ir.proto` | The IR message schema: values, matchers, detections, conditions, rule/correlation/filter, metadata, and the `Pack` envelope. |
+| `sigma_backend.proto` | The `SigmaBackend` gRPC service (`Capabilities` + `Convert`). A thin transport over the message schema; a backend can run as a local subprocess or a remote endpoint. |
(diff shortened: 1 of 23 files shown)


mostafa:
This is superseded by [Sigma CLI delegation](https://timescale.github.io/rsigma/reference/backends/sigma-cli/).

PR #205, opened 2026-06-13
feat(cli,fibratus): one rule file per rule for directory output

mostafa:
## Summary

Addresses PR review feedback that the Fibratus backend should emit a separate YAML file per Sigma rule rather than one combined `---`-joined stream, so converted rules drop straight into a Fibratus `Rules/` directory.

- `backend convert` now writes one file per converted rule when `--output` points at a directory: an existing directory, or a path with a trailing separator that is created on demand. Single-file and stdout output are unchanged (passing a directory to `-o` previously just errored on write, so this is purely additive).
- Files are named after a snake_case slug of the rule title (`Detect Whoami` -> `detect_whoami.yml`), falling back to the rule id and then a `rule` literal, with numeric-suffix de-duping so no rule overwrites another. A rule that converts to several documents (for example a `temporal` correlation expanded with `-O temporal_permute=true`) keeps them together in its one file, finalized through the backend so the format-aware separators land inside.
- New `Backend::output_file_extension(format)` hook picks the per-rule extension: `yml` for the Fibratus YAML envelope (`txt` for its bare-expression `expr` format), `sql` for PostgreSQL, `txt` by default.

```bash
rsigma backend convert rules/windows/ -t fibratus -p fibratus_windows -o ./Rules/
```

Docs updated: Fibratus backend reference, rule-conversion guide, README, the adding-backends developer note, and a CHANGELOG entry.

## Test plan

- [x] Unit tests for `slugify`, `rule_filename` (title/id/`rule` fallbacks, collision de-dupe), and directory-target detection.
- [x] CLI integration tests: one file per rule into a directory; directory created from a trailing separator.
- [x] `cargo fmt --all -- --check`, `cargo clippy` with `-Dwarnings`, full test suites for both crates (628 passed).
- [x] `mkdocs build --strict`.
- [x] Manual end-to-end run confirming file naming and content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to RSigma are documented in this file. Each entry correspond
 
 ## [Unreleased]
 
-### `backend convert`: per-rule file output when `--output` is a directory
+### `backend convert`: per-rule file output when `--output` is a directory (#205)
 
 `rsigma backend convert` can now write one file per converted rule instead of a single concatenated stream. When `--output` points at a directory (an existing directory, or a path with a trailing separator that is created on demand), each rule is written to its own file named after a snake_case slug of the rule title, with the backend's native extension. This was prompted by Fibratus rule-deployment ergonomics: the engine loads one YAML rule per file from its `Rules/` directory, so the split output drops straight in without hand-separating the `---`-joined stream.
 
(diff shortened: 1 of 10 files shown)


mostafa:
@rabbitstack Please review. 🙏

rabbitstack (approved):
LGTM!

PR #213, opened 2026-06-15
feat(rstix): add STIX meta objects

SecurityEnthusiast:
## Summary

Phase 2 — STIX meta objects (`model::meta`):

- `MarkingDefinition` (legacy TLP v1 + current TLP v2 encodings on STIX 2.1 objects, `IS_NON_VERSIONABLE`, nine TLP UUID constants, STIX §7.2.1 optional common properties)
- `ExtensionDefinition` (`created_by_ref` required per STIX §7.2.2)
- `LanguageContent` (`contents` as nested `BTreeMap` for stable JSON key order)
- `MetaObject` enum (`#[non_exhaustive]`)

Not releasable on its own until `StixObject` dispatch and `Bundle` parsing land.

## Review follow-up (second pass)

Addresses inline review on marking-def round-trip fidelity, `LanguageContent` determinism, meta `type` validation, `IS_NON_VERSIONABLE`, and round-trip helper documentation.

| Review item | Change |
|-------------|--------|
| `MarkingDefinition` missing §7.2.1 common fields | Added `created_by_ref`, `external_references`, `object_marking_refs`, `granular_markings`; rich fixture `marking-definition-with-common-props-stix21.json` |
| `LanguageContent.contents` `HashMap` | Nested `BTreeMap` |
| Meta `type` not validated | Single-pass `deserialize_with` + `ModelError::UnexpectedObjectType`; cross-type reject tests |
| `is_non_versionable()` always true | `IS_NON_VERSIONABLE` const; method delegates to it |
| `roundtrip.rs` comment overstated guarantees | Reviewer-accurate subset docs; added `roundtrip_strict` for complete types (`ExternalReference`, `GranularMarking`, `ExtensionMap`, all meta objects) |

## Testing

Two-layer layout (same as `model::common`):

| Layer | Location | Purpose |
| ----- | -------- | ------- |
| Wire / JSON | `tests/spec.rs` + `tests/fixtures/spec/meta/` | `roundtrip_strict` for complete types; subset `roundtrip` for common-property-only structs |
| Unit | `#[cfg(test)]` in `src/` | Invariants, TLP id pins, wrong-`type` rejects |

STIX object model version vs TLP encoding version is documented in `crates/rstix/README.md` and `docs/library/rstix.md`.

## Test plan

- [x] `cargo clippy -p rstix --all-targets --all-features -- -D warnings`
- [x] `cargo test -p rstix --all-features` (71 tests)
- [x] CI green on ubuntu / macOS / Windows
- [x] Reviewer: re-check marking-def rich fixture round-trip
- [x] Reviewer: confirm meta `type` rejection on cross-type fixtures


crates/rstix/README.md
```diff
@@ -75,10 +72,147 @@ assert!(json.contains("\"spec_version\":\"2.1\""));
...
+Full rationale: `crates/rstix/README.md` (Development Notes) and `docs/library/rstix.md` (Testing).
+
+## Test plan
+
+- [ ] `cargo fmt --all -- --check`
+- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
+- [ ] `cargo test -p rstix --all-features`
+```
+
+
```
mostafa: Remove this section:
```suggestion
```

crates/rstix/src/model/meta/language_content.rs
```diff
@@ -0,0 +1,89 @@
...
+///
+/// let json = r#"{
+///   "type": "language-content",
+///   "spec_version": "2.1",
+///   "id": "language-content--a0b0c0d0-e0f0-4a5b-8c9d-0e1f2a3b4c5d",
+///   "created": "2016-05-12T08:17:27.000Z",
+///   "modified": "2016-05-12T08:17:27.000Z",
+///   "object_ref": "attack-pattern--0c7b5b88-8ff7-4a4d-aa9d-feb398cd0061",
+///   "object_modified": false,
+///   "contents": { "de": { "name": "Speerphishing" } }
```
mostafa: Maybe?
```suggestion
///   "contents": { "de": { "name": "Spear-phishing" } }
```

mostafa (approved):

crates/rstix/README.md
SecurityEnthusiast: Fixed.

crates/rstix/src/model/meta/language_content.rs
SecurityEnthusiast: Both misspelling in doctest and language-content.json corrected.

crates/rstix/src/model/meta/marking_def.rs
```diff
@@ -0,0 +1,216 @@
...
+        serde(default, skip_serializing_if = "Option::is_none")
+    )]
+    pub name: Option<String>,
+    /// TLP 2.0 and other extension payloads (current encoding for predefined TLP markings).
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "ExtensionMap::is_empty")
+    )]
+    pub extensions: ExtensionMap,
+}
```
mostafa, line 157: Should fix: `MarkingDefinition` silently drops spec-valid common properties on round-trip. Per STIX 2.1 §7.2.1 and the common-properties applicability table, a `marking-definition` may carry `created_by_ref`, `external_references`, `object_marking_refs`, and `granular_markings` (all optional — a marking can itself be marked, and `created_by_ref` is common in practice). None are modeled here, and with derived serde, no `deny_unknown_fields`, and no flatten-capture, any marking carrying those fields loses them on re-serialize.

The `roundtrip` subset-compare won't catch this because it only checks that emitted fields match the fixture, not that fixture fields survive, and the TLP fixtures are minimal. Suggest modeling the four optional fields (mirroring the `Option`/`Vec` + `skip_serializing_if` pattern already used here), or capturing unknowns into a `BTreeMap` flatten field for lossless round-trip.

crates/rstix/src/model/meta/language_content.rs
```diff
@@ -0,0 +1,89 @@
...
+    object_type: String,
+    /// SDO/SRO common properties.
+    #[cfg_attr(feature = "serde", serde(flatten))]
+    pub common: SdoSroCommonProps,
+    /// Target object id.
+    pub object_ref: StixId,
+    /// Whether `object_ref` points at the latest revision of the target object.
+    pub object_modified: bool,
+    /// Translations keyed by language tag, then by field name.
+    pub contents: HashMap<String, HashMap<String, serde_json::Value>>,
```
mostafa: `HashMap` serializes in arbitrary order, so `contents` emits non-deterministic JSON. The rest of the crate uses `BTreeMap` for stable output (`ExtensionMap`, `ExternalReference.hashes`), and the crate depends on `serde_jcs` for canonical JSON, so this is inconsistent. The single-key fixture hides it today, but a document with 2+ languages or fields can produce unstable output.

```suggestion
    pub contents: std::collections::BTreeMap<String, std::collections::BTreeMap<String, serde_json::Value>>,
```

(and drop the `use std::collections::HashMap;` import at the top).

mostafa (commented):
Follow-up deeper pass on the meta objects (on top of the earlier approval). I re-ran the gates locally on this head: `cargo fmt --check`, `cargo check -p rstix --no-default-features`, `cargo clippy -p rstix --all-targets --all-features -- -D warnings`, and `cargo test -p rstix --all-features` (67 tests) all pass.

Nice work overall: the TLP UUID constants pinned against spec literals, the legacy-v1 vs current-v2 encoding split, the `created_by_ref` enforcement with both a unit test and a reject fixture, and moving pure-construction tests out of `spec.rs` into `src` unit modules all look great.

Inline: one substantive round-trip lossiness item on `MarkingDefinition`, one determinism item on `LanguageContent.contents`, a type-validation decision, and two minor notes.

crates/rstix/src/model/meta/extension_def.rs
```diff
@@ -0,0 +1,183 @@
...
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for ExtensionDefinition {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        #[derive(serde::Deserialize)]
+        struct Raw {
+            #[serde(rename = "type", default = "extension_definition_type")]
+            object_type: String,
```
mostafa, line 126: Decision (non-blocking): the `type` field is captured but never validated against `TYPE_NAME` on any of the three meta types, and with no `deny_unknown_fields` a `language-content` document deserializes cleanly as a `MarkingDefinition` (silent type confusion). Fine if the future `StixObject` dispatch is meant to be the sole gate, but a cheap `if raw.object_type != Self::TYPE_NAME { return Err(de::Error::custom(...)) }` here (and in the derived types) would prevent mis-parsing now. Your call whether to add it in this slice or defer to the dispatch PR.

crates/rstix/src/model/meta/marking_def.rs
```diff
@@ -0,0 +1,216 @@
...
+
+impl MarkingDefinition {
+    /// STIX type name for marking definitions.
+    pub const TYPE_NAME: &'static str = "marking-definition";
+
+    /// Marking definitions must not be updated after creation (STIX §7.2.1).
+    pub const fn is_non_versionable(&self) -> bool {
+        true
+    }
```
mostafa: Minor: `is_non_versionable(&self)` always returns `true` and ignores `self`, which reads as if it could be conditional. An associated `const IS_NON_VERSIONABLE: bool = true;` (or keeping the method but documenting that it's invariant for all marking definitions) is more honest. `clippy::unused_self` is pedantic-only so CI won't flag it.

crates/rstix/tests/support/roundtrip.rs
```diff
@@ -38,8 +38,8 @@ pub fn assert_fixture_rejects<T: DeserializeOwned>(relative_path: &str) {
 /// `name`, …) that the type deliberately ignores today; for those object
 /// fixtures every emitted field must match the fixture, but extra fixture keys
 /// are allowed. Standalone type fixtures (for example `ExternalReference`) use
-/// full value equality. Once concrete SDO/SRO types land in later slices, full
-/// fixture comparison will catch dropped fields for free.
+/// full value equality. Once concrete SDO/SRO types land in a later Phase 2
```
mostafa: Minor (carryover from #201): this still says standalone type fixtures use full value equality, but the `(Object, Object)` arm subset-compares every object fixture, including `ExternalReference`, so a dropped field would go unnoticed there too. Worth rewording so the comment doesn't overstate the guarantee.

```suggestion
/// fixtures every emitted field must match the fixture, but extra fixture keys
/// are allowed; dropped fields are NOT caught by this object-vs-object arm.
/// Only non-object fixtures use full value equality. Once concrete SDO/SRO
/// types land in a later Phase 2 milestone, full fixture comparison will catch
/// dropped fields for free.
```

crates/rstix/src/model/meta/marking_def.rs
SecurityEnthusiast: added §7.2.1 optional common fields + rich fixture; strict round-trip on complete marking-def fixtures.

crates/rstix/src/model/meta/language_content.rs
SecurityEnthusiast: contents is nested BTreeMap.

crates/rstix/src/model/meta/extension_def.rs
SecurityEnthusiast: validate type in single serde pass on all three meta types (UnexpectedObjectType).

crates/rstix/src/model/meta/marking_def.rs
SecurityEnthusiast: IS_NON_VERSIONABLE const + documented invariant method.

crates/rstix/tests/support/roundtrip.rs
SecurityEnthusiast: subset helper comment restored per review; added roundtrip_strict for complete types.

mostafa: The split into roundtrip_strict (full equality) vs roundtrip (subset), plus the reworded doc, resolves this cleanly. Bonus that the meta and common round-trip tests now run under strict comparison.

mostafa (approved):
Re-reviewed at fc7d0df. All five threads from the second pass are addressed and resolved:

- MarkingDefinition now models the four §7.2.1 optional common properties (created_by_ref, external_references, object_marking_refs, granular_markings), with a rich fixture exercised under strict (full-equality) round-trip, so dropped fields would now fail the test
- LanguageContent.contents is a nested BTreeMap for deterministic key order
- type is validated in a single serde pass across all three meta types via type_check.rs (UnexpectedObjectType), with cross-type reject tests
- IS_NON_VERSIONABLE associated const backs the documented invariant method
- the roundtrip helper is split into roundtrip_strict vs roundtrip with accurate docs, and the existing meta/common tests were migrated to strict comparison

The strict/subset split is a nicer outcome than I suggested. Verified locally on this head: cargo fmt --check, cargo check -p rstix --no-default-features, cargo clippy -p rstix --all-targets --all-features -- -D warnings, and cargo test -p rstix --all-features (71 tests) all pass. LGTM.

SecurityEnthusiast:
Thanks @mostafa

PR #219, opened 2026-06-16
fix(convert): emit the required `version` field in Fibratus rules

mostafa:
## Summary

Follow-up to #217. The Fibratus author found that converted rules fail to load because the rule envelope is missing the required top-level `version` attribute:

```
Error: invalid rule definition:
 ...
 (root): version is required
```

`version` is the rule content version (a free-form string the upstream rules library writes as semver, e.g. `1.0.1`), distinct from `min-engine-version`. Sigma has no equivalent attribute.

## Changes

- The YAML envelope now emits `version:` right after `id` for both detection and correlation rules. It is emitted regardless of `emit_metadata` since even a minimal envelope must load.
- Defaults to `1.0.0`, overridable with `-O version=<value>`.
- Docs updated: backend options table, envelope field list, and the `default`/correlation output examples.

## Test plan

- [x] `cargo test -p rsigma-convert` (envelope unit tests updated for the new field; added coverage for the configured version and the minimal-envelope case)
- [x] `cargo test -p rsigma --test cli_convert`
- [x] `mkdocs build --strict`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to RSigma are documented in this file. Each entry correspond
 
 ## [Unreleased]
 
-### Fibratus conversion: emit the required `version` field (#NNN)
+### Fibratus conversion: emit the required `version` field (#219)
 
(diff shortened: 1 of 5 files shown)


mostafa:
@rabbitstack Please verify. 🙏

rabbitstack (approved):
Thanks for fixing!

PR #220, opened 2026-06-16
feat(rstix): add STIX relationship and sighting objects

SecurityEnthusiast:
## Summary

Adds typed STIX Relationship Objects to `rstix`: `Relationship`, `Sighting`, and the `SroObject` enum, with shared `"type"` validation hoisted to `model::type_check` for reuse by meta and SRO types.

This work is **not releasable on its own** until `StixObject` dispatch and `Bundle` parsing land in a follow-up change.

## What's included

- **`Relationship`** (STIX §5.1): all common + relationship-specific properties via flattened `SdoSroCommonProps`; `relationship_type` charset and `stop_time` > `start_time` enforced at deserialize.
- **`Sighting`** (STIX §5.2): all common + sighting-specific properties; `WhereSightedRef` (identity or location) for `where_sighted_refs`; `ObservedDataId` for `observed_data_refs`; count range and `last_seen` ≥ `first_seen` enforced at deserialize.
- **`SroObject`**: `#[non_exhaustive]` enum with `QueryableStixObject` delegation.
- **Tests**: `roundtrip_strict` minimal + rich fixtures; negative fixtures for enforced invariants and cross-type `"type"` rejects.
- **Docs**: `crates/rstix/README.md` invariant table, `docs/library/rstix.md`, `CHANGELOG.md` [Unreleased].

## Known gaps and intentional deferrals

Fields are spec-complete for both types (optional on the wire is modeled as `Option` / empty collections; `defanged` correctly omitted as N/A for SROs). The items below are **validation or integration** gaps, not missing properties.

### Reference target rules (deferred)

| Property | Spec intent | Today | Follow-up |
|----------|-------------|--------|-----------|
| `Relationship.source_ref` | Must reference SDO or SCO only | Accepts any valid `StixId` | Enforce when bundle-level typed parsing exists |
| `Relationship.target_ref` | Same | Same | Same |
| `Sighting.sighting_of_ref` | Must reference SDO only | Accepts any valid `StixId` | Same |

Documented in rustdoc and the crate README invariant table.

### Bundle / graph validation (not in this PR)

- References are not checked against objects present in a bundle.
- Spec-defined relationship-type vocabulary for source/target pairs is not validated (charset only; spec allows arbitrary strings).
- No graph-level reference integrity checks.

These belong with bundle parsing and the validation pipeline in later work.

### Dispatch and parsing (not in this PR)

- No `Deserialize for SroObject` / no JSON dispatch by `"type"` into the enum (same pattern as meta today: deserialize leaf types directly).
- `SroObject` is not wired into a top-level `StixObject` enum yet.
- `parse_bundle()` still returns `NotImplemented`; no `Bundle::parse()`.

### Lossless capture of unknown properties (not in this PR)

- Top-level `x_*` and other unmodeled keys on relationship/sighting objects are **ignored** on deserialize (no per-object `extra` map yet).
- Strict round-trip is guaranteed only for modeled fields.

### API / testing nuances

- `QueryableStixObject::get_field` exposes a subset of fields (not refs or common props).
- No wire tests for `SroObject` enum delegation (meta follows the same pattern).
- Negative fixtures exist for enforced invariants; missing required fields rely on serde errors without dedicated fixtures.
- `summary` omitted in JSON → `None`; re-serialize omits the key (spec default is false; wire may differ from `"summary": false` unless set explicitly).

### Typing choices (intentional)

- `source_ref`, `target_ref`, `sighting_of_ref`: `StixId` (not SDO/SCO/SDO-only newtypes) until ref-target rules above land.
- `where_sighted_refs`: `WhereSightedRef` (identity \| location), matching the property definition (not identity-only).
- `summary`: `Option<bool>` (STIX 2.1 JSON boolean, not the string type shown in some spec tables).

### Local workspace gates

- [x] `cargo test --workspace --all-features --locked` (2114 tests)
- [x] `mkdocs build --strict`

### `rstix` crate

- [x] `cargo test -p rstix --all-features` (89 tests)

### CI / review

- [ ] CI green on ubuntu / macOS / Windows
- [ ] Reviewer: confirm SRO type rejection on cross-type fixtures (`sro_types_reject_wrong_type_field`)


mostafa (commented):
Reviewed against STIX 2.1 §5.1–§5.2; ran `cargo test -p rstix --all-features --locked` (89 passed) and `cargo clippy -p rstix --all-features --all-targets -- -D warnings` (clean) on the branch.

Architectural fit is good: the SRO module mirrors `model/meta` exactly, the `model::type_check` hoist is clean, and the spec-derived invariants (charset, time ordering, count range, where-sighted typing, cross-type rejection) are all correct. Inline comments below cover a project-wide pattern question (default-on-missing `"type"` field), a couple of test tightenings, a fragile `type_name` shadow that also lives on `MetaObject`, and a few optional polish items.

Not a blocker for the PR's own scope — the deferrals listed in the description are all correctly out of scope here.

crates/rstix/src/model/sro/relationship.rs
```diff
@@ -0,0 +1,245 @@
...
+/// ```
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+pub struct Relationship {
+    /// STIX object type (`relationship`).
+    #[cfg_attr(
+        feature = "serde",
+        serde(
+            rename = "type",
+            default = "relationship_type_name",
```
mostafa: `default = "relationship_type_name"` here means a JSON object that omits `"type"` deserializes as if `"type": "relationship"` was present — the validator never sees the missing-key path. STIX requires `type`, so this is loose vs spec.

Same pattern lives on `MarkingDefinition`, `LanguageContent`, and `ExtensionDefinition`. Either drop the `default` here (a missing `type` becomes a serde error) and apply the same to the three meta types in this PR, or open a follow-up that fixes all four together. The `deserialize_with` continues to validate the present-key path either way.

Not a blocker, but Phase 2 is the right time to resolve this before any further typed objects copy the pattern.

crates/rstix/src/model/sro/sighting.rs
```diff
@@ -0,0 +1,325 @@
...
+/// ```
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+pub struct Sighting {
+    /// STIX object type (`sighting`).
+    #[cfg_attr(
+        feature = "serde",
+        serde(
+            rename = "type",
+            default = "sighting_type_name",
```
mostafa: Same `default = "sighting_type_name"` issue as `relationship.rs:43` — a missing `"type"` in the JSON quietly defaults instead of failing. Resolve together with `Relationship` and the three meta types.

crates/rstix/src/model/sro/relationship.rs
```diff
@@ -0,0 +1,245 @@
...
+
+#[cfg(all(test, feature = "serde"))]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn rejects_wrong_type_field() {
+        let json = include_str!("../../../tests/fixtures/spec/sro/sighting-minimal.json");
+        let err = serde_json::from_str::<Relationship>(json).unwrap_err();
+        assert!(err.to_string().contains("relationship"));
```
mostafa: Loose assertion. The error format from `ModelError::UnexpectedObjectType` is `expected STIX type \`X\`, got \`Y\``, so this matches the *expected* substring `relationship` even when the actual mismatch (`sighting`) is dropped from the message. A regression that drops `actual` would not break the test.

Tighten to either:

```rust
assert!(err.to_string().contains("expected STIX type `relationship`"));
assert!(err.to_string().contains("got `sighting`"));
```

or downcast the serde error and inspect the structured `ModelError::UnexpectedObjectType { expected, actual }` discriminant.

crates/rstix/src/model/sro/sighting.rs
```diff
@@ -0,0 +1,325 @@
...
+
+#[cfg(all(test, feature = "serde"))]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn rejects_wrong_type_field() {
+        let json = include_str!("../../../tests/fixtures/spec/sro/relationship.json");
+        let err = serde_json::from_str::<Sighting>(json).unwrap_err();
+        assert!(err.to_string().contains("sighting"));
```
mostafa: Same weak assertion as `relationship.rs:200` — strengthen to assert both `expected STIX type` and `got` substrings (or inspect the structured error).

crates/rstix/src/model/sro/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+impl QueryableStixObject for SroObject {
+    fn id(&self) -> &StixId {
+        match self {
+            Self::Relationship(inner) => inner.id(),
+            Self::Sighting(inner) => inner.id(),
+        }
+    }
+
+    fn type_name(&self) -> &'static str {
+        self.type_name()
```
mostafa: `fn type_name(&self) -> &'static str { self.type_name() }` resolves to the inherent method (lines 23–29), not infinitely-recursing, only because Rust's method resolution prefers inherent methods over trait methods. Remove the inherent method later and this becomes silent infinite recursion with no compile error.

Two tidy fixes:

- Inline the `match self` directly in the trait impl and drop the inherent method.
- Keep the inherent method but write `Self::type_name(self)` here, so a missing inherent fails to compile.

`MetaObject` has the same pattern. Pick one and apply to both — here or in a follow-up — before a third dispatch enum lands and copies it.

crates/rstix/src/model/sro/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// STIX SRO enum (2 variants).
+#[non_exhaustive]
+#[derive(Clone, Debug, PartialEq)]
+pub enum SroObject {
+    /// A relationship between two STIX objects.
+    Relationship(Relationship),
+    /// A sighting of an SDO.
+    Sighting(Sighting),
+}
```
mostafa, line 19: No test exercises `SroObject`'s `QueryableStixObject` impl directly — only the leaf types are covered. The five `match self`-arm dispatches would silently break under a refactor.

A short test would lock the delegation:

```rust
#[test]
fn sro_object_delegates_queryable_stix_object() {
    let raw = include_str!("../../../tests/fixtures/spec/sro/relationship.json");
    let r: Relationship = serde_json::from_str(raw).unwrap();
    let sro = SroObject::Relationship(r.clone());
    assert_eq!(QueryableStixObject::id(&sro), r.id());
    assert_eq!(QueryableStixObject::type_name(&sro), Relationship::TYPE_NAME);
    assert_eq!(sro.spec_version(), Some(SpecVersion::V2_1));
}
```

`MetaObject` has the same gap; cover both at once in a follow-up if not here.

crates/rstix/src/model/sro/relationship.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    fn get_field(&self, path: &[&str]) -> Option<QueryValue<'_>> {
+        match path {
+            ["relationship_type"] => Some(QueryValue::Str(&self.relationship_type)),
+            ["description"] => self.description.as_deref().map(QueryValue::Str),
+            ["start_time"] => self.start_time.as_ref().map(QueryValue::Timestamp),
+            ["stop_time"] => self.stop_time.as_ref().map(QueryValue::Timestamp),
+            _ => None,
+        }
```
mostafa, line 187: `source_ref` and `target_ref` are not reachable via `get_field`. Root cause is cross-cutting: `QueryValue` has no `Id(&StixId)` variant, so the same omission exists for `created_by_ref` on `MarkingDefinition` and the typed-id fields elsewhere.

Not in this PR's scope. Worth a follow-up that either adds `QueryValue::Id(&StixId)` or documents that refs are intentionally unreachable through the trait.

crates/rstix/src/model/sro/sighting.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    fn get_field(&self, path: &[&str]) -> Option<QueryValue<'_>> {
+        match path {
+            ["description"] => self.description.as_deref().map(QueryValue::Str),
+            ["first_seen"] => self.first_seen.as_ref().map(QueryValue::Timestamp),
+            ["last_seen"] => self.last_seen.as_ref().map(QueryValue::Timestamp),
+            ["count"] => self.count.map(|count| QueryValue::Int(i64::from(count))),
+            ["summary"] => self.summary.map(QueryValue::Bool),
+            _ => None,
```
mostafa, line 269: Same: `sighting_of_ref`, `observed_data_refs`, and `where_sighted_refs` are not reachable via `get_field`. Tracked together with the `Relationship.get_field` follow-up.

crates/rstix/src/model/sro/sighting.rs
```diff
@@ -0,0 +1,325 @@
...
+
+use crate::core::{
+    IdentityId, LocationId, ObservedDataId, QueryValue, QueryableStixObject, SpecVersion, StixId,
+    StixTimestamp,
+};
+use crate::model::ModelError;
+use crate::model::common::SdoSroCommonProps;
+
+/// Maximum inclusive value for [`Sighting::count`] (STIX §5.2.1).
+pub const SIGHTING_COUNT_MAX: u32 = 999_999_999;
```
mostafa: Stylistic: `Sighting::COUNT_MAX` would be more discoverable from `Sighting`'s rustdoc and group with `Sighting::TYPE_NAME` already on the type. Public re-export from `model::sro` either way. Optional.

crates/rstix/src/model/sro/relationship.rs
```diff
@@ -0,0 +1,245 @@
...
+        )
+    )]
+    object_type: String,
+    /// SDO/SRO common properties.
+    #[cfg_attr(feature = "serde", serde(flatten))]
+    pub common: SdoSroCommonProps,
+    /// Name of the relationship (for example `uses`, `related-to`).
+    pub relationship_type: String,
+    /// Source object id.
+    pub source_ref: StixId,
```
mostafa: Optional polish: when `StixObject` dispatch lands, the spec rule is "SDO or SCO" for `source_ref`/`target_ref` and "SDO only" for `sighting_of_ref`. A `pub type RelSourceRef = StixId;` (or similar) today gives the future migration a grep-able anchor and puts the intent in the type system, parallel to how `IdentityId` / `LocationId` already work. The rustdoc comment already says this; the type alias would just make it visible from the type.

crates/rstix/src/model/sro/mod.rs
SecurityEnthusiast: Added `sro_object_delegates_queryable_stix_object` in `crates/rstix/src/model/sro/mod.rs` (commit `bcf67f9f`). It wraps a fixture `Relationship` in `SroObject::Relationship` and asserts `id`, `type_name`, `spec_version`, and `get_field` delegate correctly through the enum impl. Same pattern for `MetaObject` in `meta/mod.rs`.

crates/rstix/src/model/sro/mod.rs
SecurityEnthusiast: Resolved.

crates/rstix/src/model/sro/relationship.rs
SecurityEnthusiast: Added `RelSourceRef`, `RelTargetRef`, and `SightingOfRef` as public `StixId` aliases with rustdoc noting SDO/SCO vs SDO-only rules. Used on the struct fields and re-exported from `model::sro` so the intent is visible in the type system and grep-able for the future `StixObject` migration.

SecurityEnthusiast:
## Review follow-up (`bcf67f9f`)

Changes after review — scope of the original PR is unchanged; deferrals (bundle ref validation, `StixObject` dispatch, `SroObject` JSON enum deserialize, lossless `x_*` capture) are still out of scope.

### Deserialize / `"type"` handling
- Removed silent default for JSON `"type"` on leaf SRO and `model::meta` types — missing `"type"` is a serde error
- Added `rejects_missing_type_field` unit tests on leaf types
- Cross-type `"type"` reject tests now assert both expected and actual type strings

### Querying / refs
- Added `QueryValue::Id`
- `get_field` exposes reference fields on SRO and meta types (`source_ref`, `target_ref`, `sighting_of_ref`, `created_by_ref`, `object_ref`, etc.)
- Fixed `SroObject` / `MetaObject` `QueryableStixObject` `type_name` delegation (trait vs inherent shadowing)

### Types
- Added public ref aliases: `RelSourceRef`, `RelTargetRef`, `SightingOfRef` (`StixId` aliases; ref-target validation still deferred until bundle-level typed parsing)
- Renamed count bound to `Sighting::COUNT_MAX`

### Tests / lint
- Added `sro_object_delegates_queryable_stix_object` and `meta_object_delegates_queryable_stix_object` unit tests
- Clippy `collapsible_if` fix in `validate()`

### Docs
- Updated `CHANGELOG.md` `[Unreleased]` for final PR scope

mostafa (approved):
All ten review comments resolved, with several propagated to the existing `meta` types so the project-wide patterns are now consistent.

Drive-by for a future PR (not blocking): mark `QueryValue` `#[non_exhaustive]` before the next variant lands so downstream `match` blocks don't break.

PR #248, opened 2026-06-25
feat(rstix): add STIX cyber-observable (SCO) objects

SecurityEnthusiast:
## Summary

Adding the full STIX 2.1 cyber-observable (SCO) layer to `rstix`:

- **18 SCO leaf types** under `model::sco` with `ScoCommonProps` flatten (not SDO common props), strict `"type"` deserialize via `deserialize_stix_type_field`, and `validate()` on the custom serde path.
- **`ScoObject`** enum (`#[non_exhaustive]`) delegating `QueryableStixObject`; `created()` / `modified()` always `None` for SCO arms (STIX §3.2).
- **Typed ref unions** in `ref_types.rs` for spec-constrained references (domain-name resolves, directory contains, network-traffic endpoints, email MIME raw refs).
- **12 predefined extensions** under `model::sco::extensions`, validated from parent `validate()` via `ExtensionMap` keys; extension-only wire covered in unit tests with `*-minimal.json` fixtures.
- **`ModelError`** variants for enforced SCO invariants; **`QueryValue`** marked `#[non_exhaustive]`.
- **49 JSON fixtures** under `tests/fixtures/spec/sco/` and integration coverage in `tests/spec.rs` (`roundtrip_strict` + negative fixtures per type).
- **Docs:** `crates/rstix/README.md` (invariant table, deferred scope), `docs/library/rstix.md`, and `CHANGELOG.md` [Unreleased] entry.

Serde follows post–#213/#220 patterns: no `default` on `"type"`, custom `Deserialize` + `validate()`, `roundtrip_strict` on complete types, `BTreeMap` for stable JSON where applicable.

**Still deferred** (documented in README): typed SDOs, `StixObject` dispatch, `Bundle` parsing, validation pipeline, graph/marking/store/TAXII runtime.

## Tests

- [x] `cargo test --workspace --all-features --locked` (2498 tests)
- [x] `mkdocs build --strict`

### rstix crate

- [x] `cargo test -p rstix --all-features --locked` (179 tests)

### CI / review

- [ ] CI green on ubuntu / macOS / Windows


crates/rstix/src/model/sco/file.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    pub fn validate(&self) -> Result<(), ModelError> {
+        if self.hashes.is_empty() && self.name.as_ref().is_none_or(String::is_empty) {
+            return Err(ModelError::FileHashesOrNameRequired);
+        }
+        if let Some(size) = self.size {
+            // size is u64 so always non-negative; no check needed
+            let _ = size;
+        }
+        ArchiveExt::validate_in_map(&self.common.extensions)?;
```
mostafa, line 93: Extension validation has no negative-path coverage anywhere in the PR. This `validate_in_map` dispatch (and the equivalent calls in `network_traffic.rs`, `process.rs`, `user_account.rs`) is only exercised by happy-path fixtures (`file-with-archive-ext.json` parses a *valid* archive-ext). The per-extension unit tests also only do `parsed.validate().expect("valid")`, and there are no `*-invalid` extension fixtures.

Result: every extension invariant (`ArchiveExtContainsRefsEmpty`, `TcpExtNoProperties`, `WindowsPeBinaryExtPeTypeEmpty`, the ~20 `*NoProperties` variants, etc.) and the parent-to-extension dispatch wiring are untested. A regression that turned any `validate_in_map` into a no-op, or any extension `validate()` into always-`Ok`, would pass CI green.

Since "12 predefined extensions, validated from parent validate() via ExtensionMap keys" is a headline feature here, please add at least: one fixture with an embedded *invalid* extension that asserts the parent `validate()` rejects it (proves the dispatch fires), plus a negative unit test per extension `validate()`. I'd treat this as the one pre-merge item.

mostafa (commented):
Reviewed against STIX 2.1 §6 (SCOs). The architecture mirrors the meta/SRO patterns from #213/#220 cleanly: no `default` on `"type"`, custom `Deserialize` + `validate()`, `roundtrip_strict` on complete types, typed ref unions with `from_stix_id` checks, `ScoObject` delegation. `QueryValue::Id` reuse and the indexed ref paths in `get_field` are a nice touch.

I independently recomputed six of the SCO ID golden vectors (artifact, email-message, software, file/SHA3 fallback, x509) with a standalone RFC-8785 + UUIDv5 script — all match, so the namespace and JCS canonicalization are correct and genuinely python-stix2-compatible rather than self-referential. (The `id/sco.rs` generator itself is pre-existing, not part of this diff; the new typed structs correctly preserve asserted IDs on parse rather than regenerating.)

One substantive gap (extension validation is untested on the negative path) plus a few minor items inline. None blocking on their own, but I'd want the test gap closed before merge since the extension validation is a headline feature of this PR.

crates/rstix/src/model/sco/process.rs
```diff
@@ -0,0 +1,248 @@
...
+            || self.created_time.is_some()
+            || self.cwd.is_some()
+            || self.command_line.is_some()
+            || !self.environment_variables.is_empty()
+            || !self.opened_connection_refs.is_empty()
+            || self.creator_user_ref.is_some()
+            || self.image_ref.is_some()
+            || self.parent_ref.is_some()
+            || !self.child_refs.is_empty()
+            || !self.common.extensions.is_empty()
```
mostafa: Counting `!self.common.extensions.is_empty()` toward "has at least one specific property" diverges from python-stix2. STIX 2.1 §6.13 says a Process MUST contain at least one property *from the process-specific list*; `extensions` is a common SCO property, and python-stix2's `_check_at_least_one_property()` excludes common properties. So a process carrying only a `windows-process-ext` and no specific property is accepted here but rejected by python-stix2.

Same pattern in `user_account.rs` and `windows_registry_key.rs`. It's arguably the more useful behavior (an extension-only object does carry content), but it's technically non-conformant and asymmetric with the reference impl. Intended? If so, worth a one-line comment in each `has_specific_property` noting the deliberate divergence; if not, drop the `extensions` term from the three checks.

crates/rstix/src/model/sco/extensions/archive.rs
```diff
@@ -0,0 +1,70 @@
...
+                obj.insert(
+                    "extension_type".into(),
+                    serde_json::Value::String(t.as_str().into()),
+                );
+            }
+            for (k, v) in &entry.properties {
+                obj.insert(k.clone(), v.clone());
+            }
+            let ext: Self = serde_json::from_value(serde_json::Value::Object(obj))
+                .map_err(|_| ModelError::ExtensionDeserializeFailed)?;
```
mostafa: `.map_err(|_| ModelError::ExtensionDeserializeFailed)` discards the underlying serde error, so a malformed embedded extension surfaces as a bare "SCO extension failed to deserialize" with no field name or reason. This pattern is duplicated verbatim across all 12 extensions. Consider carrying the serde message (e.g. an `ExtensionDeserializeFailed(String)` variant) so operators can tell *which* field of *which* extension was malformed. Minor, debuggability only.

crates/rstix/src/model/sco/network_traffic.rs
```diff
@@ -0,0 +1,307 @@
...
+        }
+        if self.is_active == Some(true) && self.end.is_some() {
+            return Err(ModelError::NetworkTrafficEndWithActive);
+        }
+        if let (Some(start), Some(end)) = (&self.start, &self.end)
+            && end < start
+        {
+            return Err(ModelError::NetworkTrafficEndBeforeStart);
+        }
+        for port in [self.src_port, self.dst_port].into_iter().flatten() {
```
mostafa: Dead no-op loop: `for port in [self.src_port, self.dst_port].into_iter().flatten() { let _ = port; }` does nothing (both are `Option<u16>`, already range-bounded). The `let _ =` is what stops clippy from flagging it. Same shape in `file.rs` (`if let Some(size) = self.size { let _ = size; }`). Drop both blocks; the comments alone document that u16/u64 can't be out of range.

crates/rstix/src/model/sco/mod.rs
```diff
@@ -0,0 +1,216 @@
+//! STIX Cyber-observable Objects (18 types).
+//!
+//! Field-level rustdoc for individual SCO properties is deferred to a follow-up
+//! documentation pass; see crate README invariant table for enforced rules.
+
+#![allow(missing_docs)]
```
mostafa, line 7: `#![allow(missing_docs)]` on the whole `sco` module (and again on `extensions/mod.rs`) plus the "field-level rustdoc deferred" note silences docs for every public field on 18 structs + 12 extensions at once. Fine as a Phase 2 expedient, but the blanket allow will also mask genuinely undocumented future additions. Worth a tracking issue so it gets narrowed (or removed) when the documentation pass lands, rather than living indefinitely.

mostafa (approved):
Thanks for your contribution! LGTM.

SecurityEnthusiast:
Thank you @mostafa

PR #265, opened 2026-06-28
feat(rstix): STIX 2.1 SDOs, bundle parse, and reference validation

SecurityEnthusiast:
## Summary

Completes the **rstix Data Model + Serialization** phase slice: typed STIX 2.1 domain objects, top-level `StixObject` dispatch, bundle ingestion with reference validation, custom-type registry, streaming parse, and fixture-backed regression coverage. This is additive library work in `crates/rstix`; no rsigma CLI/runtime behavior changes.

## Roadmap

- **Phase:** rstix — Data Model + Serialization (Phase 2 vertical slice)
- **Explicitly deferred:** Validation Pipeline semantics (granular selector paths, language-content nested rules, STIX-W0031 warnings), Pattern Engine indicator AST, graph/store/TAXII runtime

### Domain object model

- **19 SDO types** under `model::sdo/` with per-type structs, `SdoObject` enum, typed ref unions (`MalwareSampleRef`, `MalwareAnalysisSampleRef`, `SoftwareId` refs), `IndicatorPattern` enum, and `ObservedDataForm` enum (object_refs XOR deprecated embedded objects).
- **`StixObject`** top-level enum (SDO / SCO / SRO / Meta / Custom) in `model/stix_object.rs` with `QueryableStixObject` delegation and strict `"type"` deserialize on every leaf type.
- **SRO enhancements:** relationship matrix validation (55 STIX 2.1 forward entries + common types), sighting summary wire normalization, `WhereSightedRef`, count/time-window rules.
- **SCO / meta touch-ups:** ref-kind and format checks on selected SCO types; meta-object rules (marking-definition, extension-definition, language-content); `KillChainPhase`; extension map predefined-key handling.

### Bundle parse and navigation

- **`Bundle::parse` / `parse_bundle()`** — typed bundle container, duplicate-id rejection, bundle-scoped reference existence + ref-kind checks, relationship endpoint matrix, `x_*` top-level property capture (`extra_properties()`), STIX §8 bundle rules (reject bundle `spec_version`, bundle id prefix).
- **`Bundle::parse_reader` / `parse_reader_with_options`** — streaming ingestion via `LimitedReader` + `serde_json::from_reader`.
- **`ParseOptions`** — `max_nesting_depth` (64), `max_string_length` (1 MiB), `max_bundle_bytes` (256 MiB), `max_object_count` (unlimited), `allow_custom`, scoped **`TypeRegistry`** with `register_custom_type::<T>()`.
- **Navigation:** `QueryableContainer`, `BundleObjectCast` trait + `get_typed::<T>()`, `objects_of_type::<T>()`, `CustomStixObject::downcast_typed()`.

### Shared validation

- **`model/validate.rs`** — centralized validators for SDO/SRO common props, ref kinds, relationship endpoints, CAPEC/CVE external refs on attack-pattern/vulnerability, SCO format checks, bundle reference graph, language-content `object_modified` match.

### Spec-audit alignment

- Removed stricter-than-spec empty-name / empty-collection checks where STIX is silent.
- Added missing ref-kind, meta-object, extension, and bundle rules documented in `crates/rstix/README.md` invariant table.
- **By design:** `course-of-action.action` remains `Option<Value>` (not a typed enum).

## Public API surface (new / extended)

| API | Purpose |
|-----|---------|
| `parse_bundle(&str)` | Parse STIX 2.1 bundle JSON |
| `model::Bundle` | Typed container + serialize + ref validation |
| `model::Bundle::parse_with_options` / `parse_reader_with_options` | Limits + custom types |
| `model::ParseOptions` / `TypeRegistry` | Parse-time configuration |
| `model::StixObject` / `SdoObject` / `ScoObject` / `SroObject` / `MetaObject` | Typed dispatch |
| `model::QueryableContainer` / `BundleObjectCast` | Ergonomic bundle navigation |
| `model::validate::*` (internal) | Shared invariant checks used at deserialize |

## Test plan

All gates run locally with `RUSTFLAGS=-Dwarnings` on macOS:

| Gate | Result |
|------|--------|
| `cargo fmt --all -- --check` | PASS |
| `cargo check --workspace --all-targets --all-features --locked` | PASS |
| `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` | PASS |
| `cargo test --workspace --all-features --locked` | PASS — **2823 tests** |
| `cargo test -p rstix --all-features --locked` | PASS — **304 tests** (5 suites) |
| MSRV `rustup run 1.88.0 cargo check ...` | PASS |
| `cargo doc --workspace --all-features --locked --no-deps` (`RUSTDOCFLAGS=-D warnings`) | PASS |
| `cargo llvm-cov --workspace --all-features --locked --summary-only` | PASS — **78.49%** line coverage |
| Sigma corpus @ `994da166…` (3132 rules) | PASS |
| Sigma + 5 dynamic pipelines + golden resolve | PASS |
| MCP smoke (stdio + HTTP) | PASS |
| `cargo audit` / `cargo deny check` | PASS |
| `zizmor --pedantic .github/workflows` | PASS |

### rstix-specific test coverage

- **`tests/spec.rs`** — wire round-trip + negative fixtures for all meta, SRO, SCO, and 19 SDO types; `roundtrip_strict` for complete types.
- **`tests/bundle.rs`** — minimal bundle, relationship ref validation, missing-ref rejection, serialize shape.
- **`tests/integration.rs`** — `TypeRegistry` round-trip, scoped registry isolation, 5k-object synthetic ATT&CK streaming via `parse_reader`, parse→serialize→reparse; optional real corpus via `RSTIX_ATTCK_BUNDLE` env var.

## Not in this PR (documented deferrals)

| Item | Phase |
|------|-------|
| Granular marking selector **semantic** validation | Validation Pipeline |
| Language-content nested rules vs target object | Validation Pipeline |
| STIX-W0031 (TLP v1 encoding warning) | Validation Pipeline |
| Location ISO 3166 code validation | Validation Pipeline |
| Standalone unknown top-level keys on leaf deserialize | Data Model polish |
| Observed-data deprecated `objects` map with SRO members | Data Model polish |
| Partial SCO `*_enc` encrypted field support | Data Model polish |
| Indicator pattern AST parse | Pattern Engine |
| Real MITRE ATT&CK corpus in CI (optional `RSTIX_ATTCK_BUNDLE`) | Ship gate polish |

## Breaking changes

None for downstream rsigma crates. `rstix` public API is extended; existing core/meta/SRO/SCO types retain prior behavior with stricter spec-aligned validation in some edge cases (documented in README invariant table).


mostafa (commented):
Reviewed in a worktree against the PR head. This is a big, well-structured slice: the `StixObject` dispatch, the bundle ref-graph validation, the 55-entry relationship matrix with the `__SCO__` sentinel, `ParseOptions`/`TypeRegistry` scoping, and the streaming `LimitedReader` are all clean and mostly spec-faithful. `cargo test -p rstix --all-features` is green (304) and `cargo fmt --check` passes.

Two things block a green merge as-is, plus a few notes inline.

1. **Clippy is red.** `cargo clippy -p rstix --all-features --all-targets --locked -- -D warnings` exits 101 on `tests/integration.rs:119` (`uninlined_format_args`). CI runs exactly this, so the gate will fail despite the PR test-plan claiming PASS (likely a local clippy/toolchain version difference).

2. **The SCO deterministic-id check rejects spec-legal input.** `verify_sco_deterministic_id` is enforced as a hard parse error, turning a STIX SHOULD into a MUST. Details inline on `sco/mod.rs`.

The rest are notes/questions, not blockers. Holding the approve until at least 1 and 2 are resolved.

crates/rstix/src/model/sco/mod.rs
```diff
@@ -188,6 +188,102 @@ impl QueryableStixObject for ScoObject {
...
+        .get("type")
+        .and_then(serde_json::Value::as_str)
+        .ok_or_else(|| {
+            serde_json::Error::io(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "SCO object missing type field",
+            ))
+        })?;
+
+    crate::model::validate::validate_sco_deterministic_id(type_name, &value)
```
mostafa: This makes SCO id verification a hard parse error for every SCO that has id-contributing properties, and there is no `ParseOptions` opt-out. That over-enforces the spec.

STIX 2.1 §2.9 makes deterministic UUIDv5 a **SHOULD**, not a MUST: a producer that can't or won't follow it MAY emit a UUIDv4. `verify_sco_deterministic_id`'s own doc comment even says "(STIX §2.9 / §3.4 SHOULD)", but the call here rejects on mismatch. Consequences:

- A spec-legal `file`/`domain-name`/etc. carrying a UUIDv4 id (contributing props present) fails to parse the entire bundle.
- Any feed whose contributing-property selection differs from this implementation by even one field (hash-algorithm precedence, an optional prop python-stix2 includes/excludes) is rejected, even though its ids are valid.
- It contradicts the preserve-asserted-id design from the SCO PR (#248 / the StoredSco model): asserted ids are authoritative; a recomputed fingerprint is for dedup *reporting*, never for rejection. python-stix2 itself does not re-verify ids on parse.

This belongs in the Validation Pipeline phase as a Warning-level diagnostic (same bucket as the STIX-W0031 / selector-semantics items already deferred in the PR description), not as a Data-Model parse rejection. Suggest dropping the call from `deserialize_sco_object_from_value` for now, or gating it behind an explicit opt-in `ParseOptions` flag that defaults off. There's also currently no test for the legal-divergence case (a UUIDv4 SCO with contributing props should parse), which is why this slipped through — every fixture happens to use the deterministic id.

crates/rstix/tests/integration.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    fn synthetic_large_bundle(object_count: usize) -> Vec<u8> {
+        let mut json = String::from(
+            r#"{"type":"bundle","id":"bundle--00000000-0000-0000-0000-000000000099","objects":["#,
+        );
+        for index in 0..object_count {
+            if index > 0 {
+                json.push(',');
+            }
+            json.push_str(&format!(
```
mostafa, line 119: `cargo clippy --all-targets -- -D warnings` fails here with `clippy::uninlined_format_args` (verified: clippy exits 101). The `tail = index` binding and any remaining positional args should be inlined into the format string (or the few hex-formatted args restructured) so the lint clears. This is the CI clippy gate, so it has to be green before merge.

crates/rstix/src/model/bundle.rs
```diff
@@ -0,0 +1,491 @@
...
+}
+
+impl Bundle {
+    /// Parse a bundle using default [`ParseOptions`].
+    pub fn parse(json: &str) -> Result<Self, crate::ParseError> {
+        Self::parse_with_options(json, &ParseOptions::default())
+    }
+
+    /// Parse a bundle with explicit options.
+    pub fn parse_with_options(json: &str, opts: &ParseOptions) -> Result<Self, crate::ParseError> {
```
mostafa, line 60: Asymmetric DoS guard: `parse_with_options` (and therefore the headline `parse_bundle(&str)`) calls `serde_json::from_str` on the full input with no byte cap. `max_bundle_bytes` is only enforced on the `parse_reader_*` path via `LimitedReader`. Combined with the default `max_object_count = usize::MAX`, the most common entry point parses an arbitrarily large untrusted string into a `Value` with only the depth/string-length guards applied.

Either add a `if json.len() > opts.max_bundle_bytes { return Err(BundleByteLimitExceeded ...) }` at the top of `parse_with_options`, or document clearly that untrusted input must go through `parse_reader`. Right now the safer path is the less obvious one.

crates/rstix/src/model/stix_object.rs
```diff
@@ -0,0 +1,653 @@
...
+        let is_toplevel = entry_obj
+            .get("extension_type")
+            .and_then(serde_json::Value::as_str)
+            .and_then(ExtensionType::from_str_value)
+            .is_some_and(|kind| kind == ExtensionType::ToplevelPropertyExtension);
+        if !is_toplevel {
+            ext_map.insert(key, entry);
+            continue;
+        }
+        for (prop, value) in entry_obj {
```
mostafa, line 476: `peel_toplevel_property_extensions` hoists a toplevel-property-extension's props to the object's top level, but the typed SDO/SCO structs don't `deny_unknown_fields` and have no flatten-catchall, so those hoisted keys are dropped during typed deserialize. They're also not captured into `extra_properties` (the `x_*` peel ran earlier), so `Bundle::serialize` won't re-emit them. Net effect: a bundle containing a toplevel-property-extension does not round-trip losslessly.

This is arguably covered by the documented "standalone unknown top-level keys on leaf deserialize" deferral, so not a blocker, but it's worth either a round-trip test that pins the current (lossy) behavior explicitly or capturing the hoisted props into `extra_properties` so they survive. Right now the loss is silent.

crates/rstix/src/model/bundle.rs
```diff
@@ -0,0 +1,491 @@
...
+            });
+        }
+
+        let mut objects = Vec::with_capacity(object_values.len());
+        let mut id_index = HashMap::with_capacity(object_values.len());
+        let mut raw_objects = HashMap::with_capacity(object_values.len());
+        let mut extra_properties = HashMap::with_capacity(object_values.len());
+
+        for value in object_values {
+            validate_value_limits(&value, opts)?;
```
mostafa: Minor: `validate_value_limits(&root, ...)` at line 93 already recurses through the entire tree (including every object), and this re-runs the full depth+string-length walk per object, so nested object content is traversed twice. Also note the depth guard runs post-parse — serde_json's own recursion limit (128) is what actually protects the parse from stack overflow, so `max_nesting_depth` is effectively a secondary, stricter cap rather than the primary guard. Neither is a correctness problem; just redundant work worth trimming.

crates/rstix/src/model/sco/mod.rs
SecurityEnthusiast: Agreed. Treating §2.9 as a MUST at parse time was wrong for ingest.

Removed the `validate_sco_deterministic_id` call from `deserialize_sco_object_from_value`. The helper remains in `validate.rs` (documented as Validation Pipeline / optional diagnostic use). Added `sco_uuid4_with_contributing_properties_parses` in `tests/bundle.rs`: a `file` with `name` and a UUIDv4 id now parses successfully.

Good to gate this behind an opt-in `ParseOptions` flag later if you prefer explicit strict mode before the full Validation Pipeline lands.

crates/rstix/tests/integration.rs
SecurityEnthusiast: Fixed. Inlined `{index}` in the format string so `clippy::uninlined_format_args` clears under `-D warnings`. Verified with `cargo clippy -p rstix --all-targets --all-features -- -D warnings`.

crates/rstix/src/model/bundle.rs
SecurityEnthusiast: NIce catch. Added an upfront `json.len() > opts.max_bundle_bytes` check in `parse_with_options` (returns `BundleByteLimitExceeded`) so the string and reader paths share the same limit. Test: `parse_with_options_enforces_max_bundle_bytes`.

Longer term I'd like to delegate string parse to `parse_reader(Cursor::new(...))` for a single enforcement path.

crates/rstix/src/model/stix_object.rs
SecurityEnthusiast: Addressed lossless preservation for hoisted extension props:

- `peel_toplevel_property_extensions` now returns hoisted keys; they are merged into `extra_properties` (in addition to being hoisted for typed deserialize when applicable).
- `capture_unmodeled_properties` picks up top-level keys that typed serialize omits (covers reparse after `Bundle` merges `extra_properties` back onto the wire).

Added `toplevel_property_extension_round_trips_via_extra_properties` in `tests/bundle.rs`.

A fuller `raw_objects`-based serialize path is the next step; I'll propose that separately so we can define merge semantics without blocking this PR.

crates/rstix/src/model/bundle.rs
SecurityEnthusiast: Removed the per-object `validate_value_limits` call. The root walk at the start of `parse_root_value` already traverses the full tree (including every bundle object). Agreed it's redundant work.

Streaming/pre-parse limits can be a separate hardening item for a later phase.

SecurityEnthusiast:
Opened a non-blocking issue to seek @mostafa direction on three possible follow-ups after this PR (parse limits, validation warnings, lossless serialize): https://github.com/timescale/rsigma/issues/267

mostafa (approved):
Thank you for your contribution! LGTM.

PR #268, opened 2026-06-29
feat(rstix): complete Data Model + Serialization with semantic validation

SecurityEnthusiast:
## Summary

Closes the **rstix Data Model + Serialization** phase. Typed STIX 2.1 objects, bundle parse/stream, extension round-trip, and bundle-scoped MUST ref checks were delivered in #265; this PR finishes the remaining spec-audit items and adds the semantic validation layer.

**Data modeling is complete** for Phase 2 scope: all 42 built-in STIX types, `Bundle::parse` / `parse_reader`, `TypeRegistry`, `x_*` / extension round-trip, and `Bundle::validate()` for SHOULD-level advisory rules. Parse stays permissive for non-MUST violations (maintainer #267 direction).

### Highlights

- **`Bundle::validate()`** — `ValidationReport` / `ValidationCode` warnings: STIX-W0031 (TLP v1), SCO deterministic id mismatch, granular-marking selector semantics, language-content field/type/list-length checks, ISO 3166 country, region open vocab, CAPEC/CVE external refs, relationship endpoint matrix, encryption-algorithm closed vocab
- **`common.extra`** on SDO/SRO/SCO common props + `MarkingDefinition.extra` — unknown keys drained into `extra_properties` on bundle parse
- **`ObservedDataEmbeddedObject`** — deprecated `objects` map accepts embedded SCO or SRO
- **`email-message`** — `subject_enc`, `body_enc` (STIX §6.6.2)
- **Removed `Bundle::raw_object()`** — round-trip via `extra_properties` only
- **Docs** — `docs/library/rstix.md`, crate README, architecture, feature flags, workspace index

**Not in scope (later phases):** Pattern Engine, full 12-phase Validation Pipeline, graph/marking/store, TAXII client, enrichment/testing harness .

## Local test plan (all passed)

Environment: macOS, `RUSTFLAGS=-Dwarnings`, workspace `--all-features --locked`.

### Core CI gates

| Gate | Command | Result |
|------|---------|--------|
| Format | `cargo fmt --all -- --check` | Pass |
| Check | `cargo check --workspace --all-targets --all-features --locked` | Pass |
| MSRV (1.88.0) | `cargo +1.88.0 check --workspace --all-targets --all-features --locked` | Pass |
| Clippy | `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` | Pass |
| Test | `cargo test --workspace --all-features --locked` | **2893 passed** (79 suites) |
[...]

PR #269, opened 2026-06-29
fix(rstix): PR #268 review follow-up (docs + validation tests)

SecurityEnthusiast:
## Summary

Follow-up to merged PR #268 (@mostafa review). Closes non-blocking items #1–#3; defers #4 (validate performance / wire cache).

### Docs
- observed-data embedded SRO intent (STIX 2.1 §4.14.1)
- language-content `object_modified` checked at validate time (§7.1.1), not parse

### Tests
**Reviewer-requested (#3):** integration fixtures for validation codes called out in review:
- `ScoDeterministicIdMismatch`
- `LanguageContentFieldUnknown`
- `StixW0031TlpV1Encoding` (via `object_marking_refs`)

**Additional:** `LanguageContentObjectModifiedMismatch` — pairs with the #2 doc change so the validate-time `object_modified` check has explicit test coverage, not only rustdoc.

### Deferred (#4)
`validate()` re-serializes each object to JSON for selector / SCO-id / language-content checks (see `validation.rs` — `to_value` in the main loop and again for targets). Reviewer suggested caching wire JSON on `Bundle` to avoid redundant work at ATT&CK scale.

**Out of scope for this PR** (docs + test fixtures only). 
**Follow-up:** dedicated perf PR with lazy wire cache; behavior and advisory validate API unchanged.

## Test plan

- [x] `cargo test -p rstix --features serde --test validation` — 15 passed
- [ ] CI green on all OS

### DIFF
--- a/crates/rstix/tests/fixtures/validation/bundle-tlp1-marking-ref.json
+++ b/crates/rstix/tests/fixtures/validation/bundle-tlp1-marking-ref.json
@@ -7,11 +7,7 @@
       "spec_version": "2.1",
       "id": "marking-definition--613f2e26-407d-48c7-9eca-b8e91df99dc9",
       "created": "2017-01-20T00:00:00.000Z",
-      "definition_type": "tlp",
-      "name": "TLP:WHITE",
-      "definition": {
-        "tlp": "white"
-      }
+      "name": "TLP:WHITE"
     },
     {
       "type": "attack-pattern",
--- a/crates/rstix/tests/fixtures/validation/bundle-tlp1-marking-ref.json
+++ b/crates/rstix/tests/fixtures/validation/bundle-tlp1-marking-ref.json
@@ -1,0 +1,28 @@
+{
+  "type": "bundle",
+  "id": "bundle--00000000-0000-0000-0000-00000000001a",
+  "objects": [
+    {
+      "type": "marking-definition",
+      "spec_version": "2.1",
+      "id": "marking-definition--613f2e26-407d-48c7-9eca-b8e91df99dc9",
+      "created": "2017-01-20T00:00:00.000Z",
+      "definition_type": "tlp",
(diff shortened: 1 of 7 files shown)


crates/rstix/tests/fixtures/validation/bundle-tlp1-marking-ref.json
```diff
@@ -0,0 +1,28 @@
+{
+  "type": "bundle",
+  "id": "bundle--00000000-0000-0000-0000-00000000001a",
+  "objects": [
+    {
```
mostafa, line 5: Trivial nit (non-blocking): this fixture has both a TLP1 marking-definition with `definition_type` and an attack-pattern referencing it via `object_marking_refs`, so both independently emit `StixW0031TlpV1Encoding`. That means `tlp1_object_marking_ref_warns` would still pass even if the `object_marking_refs` detection path (`check_tlp_v1_usage`) broke, so it doesn't isolate the path its name implies. Dropping `definition_type`/`definition` from the marking-definition here (keeping just the predefined TLP1 id) would make it a true guard for the ref path. Tiny, fine to leave.

mostafa (approved):
Approving. Tidy follow-up that closes observations #1-#3 from the #268 review, with #4 (validate re-serialize cost) correctly deferred to a dedicated perf PR.

Verified in a worktree: validation suite 15 passed, `cargo clippy --all-features --all-targets --locked -- -D warnings` clean, `cargo fmt --check` clean. The new fixtures are genuine triggers (the deterministic-id one uses the real `example.com` vs all-zeros mismatch; the object-modified fixture sets `contents.de.name` to match the target so only the intended code fires). Docs cite §4.14.1 and §7.1.1 for the two intent clarifications.

One trivial non-blocking nit inline.

mostafa (approved):
Thank you for your contribution! LGTM.

PR #272, opened 2026-06-29
feat(rstix): Pattern Engine: lexer, Level 1–3 parser, and type-checker

SecurityEnthusiast:
## Summary

First shippable slice of the **rstix Pattern Engine**: STIX Specification **§9** patterning with **parse + type-check only**. Adds the optional `pattern` feature, hand-written lexer, recursive-descent parser (Levels 1–3), SCO schema type-checker, §9.8 fixture-backed tests, and documentation on the [rstix library page](docs/library/rstix.md#pattern-engine-stix-9) and [crate README](crates/rstix/README.md#pattern-engine-stix-9).

**Evaluation**, **canonical printer**, **`IndicatorPattern::Stix { ast }` serde**, and **fuzz** are intentionally **not** in this PR (documented as deferred follow-on Pattern Engine work). Grammar authority: **STIX §9** (not §8).

---

### Feature flag and wiring

- Cargo feature **`pattern`** → optional `base64` (binary literals).
- `[[test]] pattern_parse` with `required-features = ["pattern"]`.
- Exports: `Pattern`, `PatternAst`, `PatternScoType`, `PatternError`, `PatternMatchError`.

### Modules (`crates/rstix/src/pattern/`)

| Module | Role |
|--------|------|
| `lexer.rs` | Tokenizer; 64 KiB input cap |
| `parser.rs` | Levels 1–3: observations, `AND`/`OR`/`FOLLOWEDBY`, `WITHIN`/`REPEATS`/`START`/`STOP` |
| `ast.rs` | `PatternAst`, `PatternScoType` (built-in + custom SCO types), paths, constants |
| `typeck.rs` | 18 SCO schemas, `_ref` chains, extensions, operator/constant validation |
| `error.rs` | `PatternError`; `PatternMatchError` (defined for future eval) |
| `mod.rs` | `Pattern::parse`, `parse_level1`, `observed_types()`, `observed_type_names()` |

### Grammar

**Level 1:** `[ … ]` with `=`, `!=`, `>`, `<`, `>=`, `<=`, `IN`, `LIKE`, `MATCHES`, `ISSUBSET`, `ISSUPERSET`, `EXISTS`, `NOT` modifier; object paths (dict keys, `hashes.MD5` dot syntax, `[*]`, `_ref` chains); string/int/float/bool/timestamp/hex/binary constants.

**Level 2:** `AND`, `OR`, `FOLLOWEDBY` between observations.

**Level 3:** `WITHIN`, `REPEATS`, `START`/`STOP`.

[...]

PR #276, opened 2026-07-01
feat(rstix): add STIX pattern evaluation (Levels 1–3)

SecurityEnthusiast:
## Summary

Adds full STIX §9 pattern **evaluation** to the `pattern` feature, building on #272.

### Evaluation engine

- **`Pattern::evaluate`** — Levels 2–3: `AND`, `OR`, `FOLLOWEDBY`, `WITHIN`, `REPEATS`, `START`/`STOP`
- **`Pattern::matches_single`** / **`Pattern::matches_single_with_bundle`** — Level 1 with optional bundle for `_ref` dereference
- **`Pattern::evaluate_observed_data`** — context from `observed-data.object_refs`
- **`ObservationContext`** / **`TimestampedObservation`** — `at: Option<StixTimestamp>`; temporal patterns return `MissingTimestamp` when timestamps are absent
- **`pattern/path.rs`** — extensions, ref lists `[*]`, nested/custom props, binary `payload_bin`, `_ref` chains, CIDR
- **`pattern/security.rs`** — `MATCHES` regex size limit (1 MiB); `compile_matches_regex` / `MATCHES_REGEX_SIZE_LIMIT`
- **`CustomSco`** — vendor types (e.g. `x-usb-device`) deserialize and evaluate
- Expanded **`get_field`** on SCO types for §9.8 paths

### Bug fixes

- **`email-addr`** deserializer (`"email-addr"`, not `"email-address"`)
- **`ScoObject::parse_str`** for integration tests

### Tests

| Suite | Coverage |
|-------|----------|
| `pattern_spec_eval.rs` | STIX §9.8 Level 1 + Level 2–3 |
| `pattern_eval_operators.rs` | Every comparison operator + temporal |
| `pattern_eval_sco_fields.rs` | **276** manifest-driven SCO field cases |
| `pattern_eval_errors.rs` | All **`PatternMatchError`** variants (22 tests) |
| `pattern_eval.rs` | Observed-data + AND context |

Test helpers use per-test `#[path = "support/…"]` includes (no `#![allow(dead_code)]`).

### Documentation Updates

- `docs/library/rstix.md`
- `docs/reference/feature-flags.md`
- `docs/reference/architecture.md`
- `docs/library/index.md`
- `docs/_data/vars.yml`

### Out of scope

- Canonical printer (`print.rs`)
- `IndicatorPattern::Stix { ast }` serde wiring
- `fuzz_stix_pattern`

## Test plan

- [x] `RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- [x] `RUSTFLAGS=-Dwarnings cargo check --workspace --all-targets --all-features --locked`
- [x] `RUSTDOCFLAGS="-D warnings -D rustdoc::broken-intra-doc-links" cargo doc --workspace --all-features --locked --no-deps`
- [x] `cargo test -p rstix --features pattern,serde` — **447 passed** (12 suites)
- [x] `cargo test --workspace --all-features --locked` — **3067 passed** (90 suites)


crates/rstix/src/pattern/path.rs
```diff
@@ -0,0 +1,1034 @@
...
+        }
+    }
+    if name == "values" {
+        if let ScoObject::WindowsRegistryKey(key) = sco {
+            if !key.values.is_empty() {
+                return Ok(vec![FieldValue::Bool(true)]);
+            }
+        }
+    }
+    if name == "extensions" {
```
mostafa: **Clippy fails the CI gate** (`cargo clippy --all-features --all-targets --locked -- -D warnings` exits 101 with 5 errors). The PR test-plan lists clippy as passing, but the pinned toolchain flags:

- `path.rs:597` and `:602` — `collapsible_if`
- `path.rs:627` — collapsible `if let`
- `path.rs:880` — needless `Ok(...?)` (`clippy::needless_question_mark`)
- `file.rs:239` — `redundant_closure`

All mechanical. This is the same local-vs-pinned clippy drift as #265; worth running `cargo clippy` on the workspace toolchain (rust-toolchain.toml pins 1.88.0) before pushing.

mostafa (commented):
Reviewed the evaluator in a worktree. A lot of this is careful and correct: the temporal semantics read right (FOLLOWEDBY `t_left <= t_right` with distinct observations, observation-level AND requiring distinct observations, WITHIN span over merged witnesses, START/STOP with exclusive stop boundary, `MissingTimestamp` when temporal ops meet timestamp-less observations), and MATCHES is protected against ReDoS via the linear `regex` crate with a 1 MiB `size_limit`. Test breadth (447 tests, manifest-driven SCO fields, all error variants) is strong.

Four things below. One blocks CI; one is a real DoS in the LIKE matcher; two are semantic/complexity notes.

crates/rstix/src/pattern/eval.rs
```diff
@@ -0,0 +1,1023 @@
...
+
+fn like_match(value: &str, pattern: &str) -> bool {
+    like_bytes(value.as_bytes(), pattern.as_bytes())
+}
+
+fn like_bytes(value: &[u8], pattern: &[u8]) -> bool {
+    if pattern.is_empty() {
+        return value.is_empty();
+    }
+    if pattern[0] == b'%' {
```
mostafa: `like_bytes` backtracks on `%` by recursing over every split point (`for i in 0..=value.len() { like_bytes(&value[i..], &pattern[1..]) }`), which is exponential (O(2^n)) on adversarial patterns like `'%a%a%a%...%a'` against a long non-matching value. MATCHES is protected (linear `regex` + size limit), but LIKE has no such guard, and both operands come from the (potentially untrusted) Indicator pattern — so a single `LIKE` comparison is a DoS vector even under the 64 KiB pattern cap.

Swap to the standard iterative two-pointer wildcard match (O(n·m), constant space): track a `star_idx` / `match_idx` backtrack point instead of recursing. Same semantics for `%`/`_`, no blowup. Worth a test with a pathological `%`-heavy pattern to lock it in.

crates/rstix/src/pattern/eval.rs
```diff
@@ -0,0 +1,1023 @@
...
+    match field {
+        FieldValue::Str(s) => Some(s.as_str()),
+        _ => None,
+    }
+}
+
+fn equals_str(s: &str, right: &PatternConstant) -> bool {
+    match right {
+        PatternConstant::String(v) => {
+            if v.contains('/') && (s.contains('.') || s.contains(':')) {
```
mostafa: This makes `=` perform CIDR containment when the RHS looks like a network and the value parses as an IP, so `[ipv4-addr:value = '10.0.0.0/8']` would match any address in the range. STIX §9.6 `=` is exact string equality; subnet containment is what `ISSUBSET` / `ISSUPERSET` are for. This produces matches the spec's `=` wouldn't (a false positive relative to conformant engines). Note the START/STOP test passes via plain `s == v` because the fixture value is literally `198.51.100.0/24`, so this branch isn't actually exercised by the suite.

Suggest dropping the CIDR fallback from `equals_str` and keeping `=` exact; if there's a real reason to keep it, document the deliberate divergence and add a test that pins it.

crates/rstix/src/pattern/eval.rs
```diff
@@ -0,0 +1,1023 @@
...
+        PatternAst::Within { inner, .. } | PatternAst::Repeats { inner, .. } => {
+            collect_witnesses(inner, ctx, window_start, window_stop)
+        }
+        PatternAst::StartStop {
+            inner, start, stop, ..
+        } => collect_witnesses(inner, ctx, Some(start), Some(stop)),
+    }
+}
+
+fn merge_disjoint_witness_sets(left: &[Vec<usize>], right: &[Vec<usize>]) -> Vec<Vec<usize>> {
```
mostafa, line 436: Complexity note (secondary): `collect_witnesses` for nested AND under WITHIN/REPEATS builds the cartesian product of sub-witness sets here, so a k-way AND over an n-observation context is O(n^k) witness tuples. The pattern side is capped (MAX_OBSERVATIONS = 256), but the evaluation context (observed SCOs, e.g. from `evaluate_observed_data` over a bundle's `object_refs`) is uncapped, so a crafted bundle + `([a] AND [b] AND [c] ...) WITHIN` pattern can blow up memory/time. Consider bounding the context size or short-circuiting witness generation once a satisfying tuple is found (WITHIN only needs existence, not enumeration).

mostafa (approved):
Thank you for your contribution! LGTM. 🚀

SecurityEnthusiast:
Thanks @mostafa

PR #296, opened 2026-07-07
feat(rstix): Pattern Engine: canonical printer, Indicator wiring, IndicatorBuilder, and fuzz

SecurityEnthusiast:
## Summary

Final **Pattern Engine** slice: canonical pattern printer, `Indicator` / `IndicatorBuilder` wiring, §9.6.1 evaluation semantics, fuzz target, and documentation (including **DD-PE-001** — `IndicatorBuilder` validates at `build()`, not in setters).

- **`Pattern::canonical` / `Display`** — AST → canonical STIX pattern string; parse → print → parse preserves semantics (§9.8 fixture round-trips).
- **`IndicatorPattern::Stix { parsed }`** — STIX indicators deserialize with `Pattern::parse(raw)` when `pattern` is enabled; invalid patterns fail at deserialize time.
- **`IndicatorPattern::evaluate` / `evaluate_observed_data`** — delegate to the parsed pattern for STIX indicators; `NonStixPattern` for YARA/Snort/etc.
- **`IndicatorBuilder`** — fluent programmatic construction (`stix_pattern`, `external_pattern`, `valid_from`, kill-chain phases); STIX patterns parse and type-check at `build()` when `pattern` is enabled; runs `Indicator::validate()`.
- **Design decision DD-PE-001** — documents why `IndicatorBuilder` validates at `build()` (materialization boundary, parity with JSON deserialize, single error surface, fluent setter chain) rather than parsing in `stix_pattern()`; recorded in `docs/library/rstix.md` and `crates/rstix/README.md`.
- **`fuzz_stix_pattern`** — libFuzzer target over parse + canonical print; seeds in `fuzz/seeds/fuzz_stix_pattern/` (§9.8 fixture lines).
- **§9.6.1 semantics** — NFC normalization for `LIKE`/`MATCHES`; PCRE DOTALL for `MATCHES`; `evaluate()` accepts `at: None` for non-temporal patterns; absent optional `_ref` properties do not match comparisons and yield `false` for `EXISTS`.
- **Docs** — `crates/rstix/README.md`, `docs/library/rstix.md`, `docs/reference/architecture.md`, `docs/reference/feature-flags.md`, `CHANGELOG.md`.

Builds on #272 (parse + type-check) and #276 (evaluation).

## Design note
- **DD-PE-001:** Setters store configuration; parse + `Indicator::validate()` run at `build()` — same materialization boundary as JSON deserialize.

## Test plan

- [x] `cargo check --workspace --all-targets --all-features --locked`
- [x] MSRV 1.88.0 check
- [x] `cargo doc --workspace --all-features --locked --no-deps` with `-D warnings`
- [x] `cargo llvm-cov --workspace --all-features --locked`
- [x] `cargo +nightly fuzz build fuzz_stix_pattern`
- [x] `mkdocs build --strict`
- [x] MCP smoke (stdio + HTTP)
- [x] Sigma corpus regression + dynamic pipeline golden resolve
- [x] `cargo audit` + `cargo deny check`
- [x] `zizmor --pedantic .github/workflows`
- [ ] CI green on GitHub

## Roadmap

Pattern Engine is **complete** after merge.


crates/rstix/src/pattern/eval.rs
```diff
@@ -650,11 +655,17 @@ fn compare_values(
             _ => false,
         },
         ComparisonOp::Like => match (field_value_str(&left), right) {
-            (Some(s), PatternConstant::String(pat)) => like_match(s, pat.as_str()),
+            (Some(s), PatternConstant::String(pat)) => {
+                let pat = normalize::nfc(pat.as_str());
```
mostafa, line 660: §9.6.1 NFC is applied to opposite single operands for the two operators, so both are only half-normalized:

- **LIKE** (here): normalizes the RHS pattern (`nfc(pat)`) but not the LHS field value `s`.
- **MATCHES** (just below): normalizes the LHS value (`nfc(&s)`) but not the RHS regex `re`.

§9.6.1 requires the comparison to be performed on NFC-normalized forms of *both* operands. As written, a field value in NFD compared against an NFC LIKE pattern still fails char-by-char (the LIKE value isn't normalized), and a MATCHES regex authored with decomposed codepoints won't match an NFC haystack (the regex isn't normalized). Only affects non-ASCII content (there's an ASCII fast-path), but STIX patterns routinely carry non-ASCII (filenames, registry values, display names).

Normalize both sides for both operators: LIKE → `nfc(s)` and `nfc(pat)`; MATCHES → `nfc(s)` and `nfc(re)` before `compile_regex`. A decomposed-vs-composed round-trip test on each operator would lock it in. Worth fixing before merge since the PR explicitly claims §9.6.1 NFC for LIKE/MATCHES.

mostafa (commented):
Reviewed the final Pattern Engine slice in a worktree. Gates green (471 tests, clippy `--all-targets -D warnings` clean, fmt clean), and the design decisions are right:

- **Round-trip fidelity preserved**: `IndicatorPattern::Stix { raw, version, parsed }` keeps the original `raw` string and `Serialize` emits `raw`, not the canonical form. The canonical printer is opt-in via `Pattern::canonical()` / `Display`. No wire regression.
- **Canonical printer** is precedence-aware at both the observation and comparison levels, escapes `'`/`\` correctly, and is tested via `semantic_eq` (AST equivalence, span-insensitive) over the §9.8 fixtures.
- Feature-gating is clean (parsed field only under `pattern`; deserialize-time validation only when the feature can validate), DD-PE-001 (validate-at-build) is sound, MATCHES DOTALL is set with a test, and the fuzz target is wired with seeds.

One §9.6.1 correctness gap inline. Tiny non-blocking nit: the STIX pattern version field is named `version` on `IndicatorPattern::Stix` but `pattern_version` on `Other` (same wire field) — aligning them would remove a small readability bump. Otherwise this is a clean close-out of the Pattern Engine.

crates/rstix/src/pattern/eval.rs
SecurityEnthusiast: Fixed. Both operands NFC-normalized for LIKE and MATCHES; added round-trip tests; aligned pattern_version on Stix.

mostafa (approved):

PR #297, opened 2026-07-08
feat(rstix): add Validation Pipeline scaffold and profiles

SecurityEnthusiast:
## Summary

Adds the **Validation Pipeline** behind the optional `validate` feature (implies `serde` + `pattern`):

- **`Validator` / `ValidatorBuilder`** — four named profiles (`consumer_permissive`, `consumer_strict`, `producer_strict`, `interop_strict`) plus custom check selection
- **Structured diagnostics** — `STIX-E/W/I/H` codes, `SourceSpan` (line/column/byte offset), `Leniency` policy on `ValidationReport::is_valid()`
- **Raw JSON entry** — `validate_json_str` / `validate_json_value` return `STIX-E0001` on malformed JSON instead of panicking
- **Early checks wired** — root type discrimination (`STIX-E0002`), parse-error bridging (`STIX-E0003` for missing ids), `with_allow_custom` / `with_parse_options`
- **Check dispatcher** — all twelve validation stages registered; ten remain stubs (schema, refs, patterns, TLP, etc.)
- **DD-VP-001** — documents the boundary between advisory `Bundle::validate()` and profile-driven `Validator`
- **`fuzz_rstix_validate_json`** — libFuzzer target + weekly CI matrix entry
- **Docs** — crate README, library docs, architecture, feature flags, fuzzing guide, CHANGELOG

## What comes next

Follow-up releases will:

1. **Implement remaining checks** — schema/required fields, reference resolution, pattern parse/type-check, TLP marking, open vocabulary, cross-object semantics, and related stages
2. **Migrate overlapping rules** from `Bundle::validate()` into the pipeline (keeping advisory `Bundle::validate()` until migration is complete)
3. **Add OASIS conformance tests** and a regression corpus

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets --all-features --locked`
- [x] `cargo test --workspace --all-features --locked` (3211 passed)
- [x] `cargo test -p rstix --all-features` (497 passed)
- [x] MSRV 1.88.0: `cargo test -p rstix --features validate`
- [x] `cargo +nightly check --bin fuzz_rstix_validate_json`
- [x] CI: `mkdocs build --strict`


crates/rstix/src/validate/validator.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// JSON, type, schema, and reference checks — permissive ingest path for mixed-trust feeds.
+    pub fn consumer_permissive() -> Self {
+        Self {
+            phases: consumer_permissive_phases(),
+            leniency: Leniency::Standard,
+            parse_options: ParseOptions::default(),
+        }
+    }
+
+    /// All twelve checks — strict validation for untrusted input.
```
mostafa: This is the one thing I'd fix before merge. `consumer_strict`'s doc says "All twelve checks — strict validation for untrusted input" and `interop_strict` says "All twelve checks with zero leniency", but 10 of the 12 checks are no-op stubs (`schema`, `id_structure`, `property_types`, `open_vocabulary`, `pattern_parse`, `pattern_semantic`, `references`, `cross_object_semantic`, `extension_resolution`, `tlp_marking` — each `pub fn run(_ctx, _report) {}`). So `Validator::consumer_strict().validate_json_str(bundle).is_valid()` returns `true` for a parseable bundle that violates schema / property-type / vocab / pattern / TLP MUST rules.

For a validator that's a false-sense-of-validation footgun: a downstream consumer reads the rustdoc (not the PR description), calls `consumer_strict` on untrusted input, and gets a green result that means "2 of 12 checks passed." Two fixes, ideally both:

1. Have each stub emit an Informational diagnostic (`STIX-I…`, e.g. `E`→`I` "schema check not yet implemented") so the report self-describes the gap. Info severity won't trip even `Leniency::Zero`, so it doesn't break the `is_valid()` contract, but `report.infos()` makes the coverage visible.
2. Tone the profile rustdoc down to enumerate implemented-vs-pending checks (or add a `Validator::implemented_phases()` accessor) so the public API doesn't claim strict validation it doesn't yet perform.

Shipping a scaffold is fine; advertising it as complete strict validation in the public docs is the part to fix.

mostafa (commented):
Reviewed the Validation Pipeline scaffold in a worktree. Gates green (497 rstix tests, clippy `--all-targets -D warnings` clean, fmt clean). As a scaffold the bones are well done: `#[non_exhaustive] ValidationPhase` in correct 12-phase order, a structured `Diagnostic` with STIX-E/W/I/H codes and line/column/byte `SourceSpan`, a no-panic raw-JSON entry (`validate_json_str` → `STIX-E0001` on malformed input), `Leniency::Zero` for interop (warnings fail), `producer_strict` correctly dropping References, and the parse-error bridge. Clean API.

One thing I'd want resolved before this lands, plus a boundary note.

(Note: this branch is stacked on #296 — the diff includes the pattern-engine files — so it should merge after #296.)

crates/rstix/src/validate/report.rs
```diff
@@ -0,0 +1,101 @@
+//! Aggregated validation output.
+
+use super::Leniency;
+use super::diagnostic::{Diagnostic, Severity};
+
+/// Validation pipeline output (distinct from [`crate::model::ValidationReport`]).
```
mostafa: Boundary/naming note (non-blocking): there are now two `ValidationReport` types — this `crate::validate::ValidationReport` and the advisory `crate::model::ValidationReport` from #268 — plus two code enums (`DiagnosticCode` here vs `ValidationCode` there). The doc comment already flags the collision, and DD-VP-001 documents the intended boundary and the migration plan, so this is deliberate and temporary. But during the migration window two identically-named report types are a real confusion surface for consumers. Worth confirming the migration is tracked in an issue, and considering a distinguishing name (e.g. `PipelineReport`) or a `#[doc(alias)]` so imports don't silently pick the wrong one.

SecurityEnthusiast:
> Reviewed the Validation Pipeline scaffold in a worktree. Gates green (497 rstix tests, clippy `--all-targets -D warnings` clean, fmt clean). As a scaffold the bones are well done: `#[non_exhaustive] ValidationPhase` in correct 12-phase order, a structured `Diagnostic` with STIX-E/W/I/H codes and line/column/byte `SourceSpan`, a no-panic raw-JSON entry (`validate_json_str` → `STIX-E0001` on malformed input), `Leniency::Zero` for interop (warnings fail), `producer_strict` correctly dropping References, and the parse-error bridge. Clean API.
> 
> One thing I'd want resolved before this lands, plus a boundary note.
> 
> (Note: this branch is stacked on #296 — the diff includes the pattern-engine files — so it should merge after #296.)

Thanks for the thorough review — glad the scaffold structure looks solid.

Stacking: #296 is merged into main now; this PR’s diff is validation-only (35 files under validate/, docs, fuzz, CHANGELOG). No Pattern Engine paths in the diff anymore.

Blocking item: Addressed in 07091ee0 — details in the inline replies below.

Re-ran cargo clippy -p rstix --all-features and cargo test -p rstix --features validate (498 passed) after the fix.

crates/rstix/src/validate/validator.rs
SecurityEnthusiast: Agreed — advertising full strict validation while 10/12 checks were no-ops was misleading. Fixed in `07091ee0` with both suggestions:

1. **Stub visibility:** Any selected but not-yet-implemented check now emits an informational `STIX-I0020` (e.g. `"schema check not yet implemented"`). Info does not fail `is_valid()` under `Leniency::Zero`, but `report.infos()` shows the coverage gap. Tests assert 10× `I0020` for `consumer_strict` on a minimal valid bundle while `is_valid()` stays `true`.

2. **Honest public API:** Profile rustdoc now states implemented vs pending coverage. Added:
   - `Validator::implemented_phases()`
   - `implemented_phases_in_profile()`
   - `pending_phases_in_profile()`
   
   README profile table includes an **Implemented today** column.

Implemented checks remain JSON well-formedness (via `validate_json_str` → `STIX-E0001`) and type discrimination (`STIX-E0002`). The rest land in the follow-up check-implementation work.

crates/rstix/src/validate/report.rs
SecurityEnthusiast: The deliberate split is documented in **DD-VP-001**; for this PR I tightened the naming story without a breaking rename:

- Crate root already re-exports `validate::ValidationReport` as **`PipelineValidationReport`** when `validate` is enabled.
- Added `#[doc(alias = "PipelineValidationReport")]` on `validate::ValidationReport` and expanded the rustdoc to steer consumers toward the crate-root alias when both report types are in scope.
- DD-VP-001 + README note that `Bundle::validate()` migration into the pipeline is follow-up work.

A full rename to `PipelineReport` (or similar) could happen when we migrate overlapping `Bundle::validate()` findings, Happy to do that in the check-implementation PR if you prefer it sooner.

mostafa (approved):

Issue #267, opened 2026-06-28
rstix: seeking direction on post-#265 improvements (non-blocking)


SecurityEnthusiast:
## Why I'm opening this

**PR #265** adds STIX 2.1 bundle parsing to `rstix` (SDOs, `Bundle::parse`, ref validation, tests, etc.).

@mostafa reviewed it. I fixed the blockers and notes in that PR. **This issue does not block #265.**

While fixing the review, I noticed three places where my **short-term fix** works, but a **cleaner long-term design** might be better. I have a rough idea for each, not a finished design. I want @mostafa direction **before** I spend time on follow-up PRs.

---

### What the problem is
Untrusted JSON should not be unbounded. We have a limit: `max_bundle_bytes` (default 256 MiB).

### What we have today (after #265)
- **`parse_reader(...)`** — reads from a stream and stops at the byte limit ✅
- **`parse_bundle("...")` / `parse_with_options(&str)`** — checks `json.len()` before parsing ✅

So both paths enforce the limit, but **two different code paths** do it. If we change limits or error handling later, we might update one path and forget the other.

### What I'm considering later
Make string parsing call the reader path internally, e.g. `parse_reader(Cursor::new(json.as_bytes()))`, so **one implementation** enforces limits.

### What I'm asking
- Is that worth a small follow-up after #265 merges?
- Any downside (error messages, performance, behavior differences)?

---

### What the problem is
STIX says SCO IDs *should* often be deterministic (UUIDv5 from object content), but **does not require it**. A UUIDv4 ID can still be valid.

### What we had in #265 (review blocker)
We **rejected** bundles when the SCO id didn't match the deterministic id — too strict.

### What we did in #265 (fix)
**Parse accepts** those objects. We kept a helper (`validate_sco_deterministic_id`) for possible future use, but parse no longer fails on mismatch.

### What I'm considering later
A **Validation Pipeline** phase (already on the roadmap) that:
- parses successfully
- then reports **warnings** like “SCO id doesn't match deterministic id” or “TLP v1 encoding used”
- without failing ingest (like python-stix2 on read)

### What I'm asking
- Is that the right split: **parse = permissive**, **validate = strict/warn**?
- Should there be an optional “strict mode” flag before the full pipeline exists, or wait?

---

### What the problem is
Real STIX bundles (e.g. MITRE ATT&CK) have extra fields and extension shapes. After **parse → serialize → parse**, we must not silently drop data.

### What we have today (after #265)
[...]

PR #315, opened 2026-07-12
feat(rstix): Validation Pipeline conformance corpus and per-diagnostic coverage

SecurityEnthusiast:
## Summary

Completes the **Validation Pipeline conformance slice** and closes remaining **phases 1–4** implementation gaps for `rstix`. Builds on prior merged work (Data Model + Serialization, Pattern Engine, Validation Pipeline check implementation).

This PR does **not** publish to crates.io.

**After merge, `rstix` with `--features validate` provides:**

- All **12 validation checks** implemented (`ValidationPhase::IMPLEMENTED`)
- **`consumer_strict`** and **`interop_strict`** run the full 12-check pipeline (other profiles run subsets **by design**)
- **39 structured diagnostics** (`STIX-E/W/I/H`) with **one integration test per code**
- **Interop-strict conformance gating** over `conformance/` + selected `validation/bundle-*.json` negatives
- **Structured diagnostics** from mapped `ModelError` / `ParseError` paths (no silent empty reports; parse limits emit `STIX-E0001`)
- **Full wire-format validators** for domain (IDNA), email (RFC 5322), and URL (WHATWG)
- **Tagged serde error propagation** (`model/serde_error.rs`): every `ModelError` at deserialize boundaries round-trips to structured diagnostics; semantic invalid bundles do not emit false `STIX-E0001`

**Marks rstix release-ready for library scope** (Core Foundation → Validation Pipeline). **Not in scope:** Graph + Marking + Store, TAXII Client, crates.io publish.

---

## Motivation

Prior Validation Pipeline work shipped all twelve checks and the diagnostic catalog, but gaps remained:

- No guarantee every `DiagnosticCode::ALL` entry was exercised through `Validator::validate_json_str`
- Conformance gating was weak (`has errors` vs `interop_strict` validity; duplicate fixture copies)
- `ModelError` was lost at serde boundaries and often surfaced as false `STIX-E0001` via display-string reverse mapping
- `ModelError` → diagnostic mapping used a catch-all `E0003` for many variants
- Wire-format validation was lightweight only (README documented IDNA/RFC 5322/WHATWG as not implemented)

This PR closes those gaps for phases 1–4.

---

### 1. Per-diagnostic integration coverage (`tests/validate_diagnostic_coverage.rs`)

- One `CoverageCase` per `DiagnosticCode::ALL` entry (**39 codes**)
- Each case runs through `Validator::consumer_strict()` or `Validator::interop_strict()` as appropriate
- Asserts the expected code appears in the pipeline report
- **Fails at test time** if `coverage_cases()` drifts from `DiagnosticCode::ALL` (missing or duplicate entries)

**Note:** `STIX-E0024` is shared by granular-selector and language-content mismatch paths. `STIX-W0010` coverage uses CAPEC; CVE is covered via `validation/bundle-bad-cve.json` in the conformance gate.

### 2. Conformance corpus gating (`tests/validate_conformance.rs`)

| Tier | Gate |
|------|------|
| `conformance/valid/` | `interop_strict` → `report.is_valid()` |
| `conformance/invalid/` + `validation/bundle-*.json` | `interop_strict` → `!report.is_valid()` |
| `conformance/versioning/` | emits `STIX-W0003` or `STIX-W0004` |
| Invalid corpus (above) | `interop_strict` → no `STIX-E0001` on semantic negatives |

**Invalid corpus sources:**

- `tests/fixtures/conformance/invalid/` (e.g. `missing_object_id.json`)
- `tests/fixtures/validation/bundle-*.json` negatives (no duplicate copies under `conformance/`)

**Excluded from invalid gate** (info-only under `interop_strict`):

- `bundle-location-bad-region.json` (`STIX-I0001`)
- `bundle-relationship-matrix-invalid.json` (`STIX-I0002`)

### 3. Exhaustive model error mapping (`src/validate/model_mapping.rs`)

New module maps every `ModelError` variant to the correct `DiagnosticCode` and default property path. `model_bridge.rs` delegates here instead of a catch-all `E0003`. The `match` in `model_mapping.rs` is exhaustive (no catch-all `_ => E0003`).

### 4. Parse bridge + serde error propagation (`parse_bridge.rs`, `stix_object.rs`, `model/serde_error.rs`, `model/error.rs`)

- All `ParseError` variants emit structured diagnostics (including object/byte/nesting/string limits)
- **`model/serde_error.rs`**: `ModelError::into_de_custom()` embeds a tagged JSON payload plus human-readable `Display` text at every serde boundary (`validate()`, type checks, ref kinds, etc.)
- Recovery: `ModelError::from_serde_message()` → tagged decode first, `from_display_message()` as legacy fallback only
- **`ParseError::Json` mapping:**
  - tagged / recovered `ModelError` → structured `STIX-E/W/I/H` via `model_mapping`
  - JSON **syntax / EOF** → `STIX-E0001`
  - other serde structural errors (e.g. missing field) → `STIX-E0003` with fix suggestion
- Parse limits (`ObjectLimitExceeded`, `BundleByteLimitExceeded`, nesting/string caps) still emit `STIX-E0001` by design

### 5. Wire-format validators (`src/model/validate.rs`, `Cargo.toml`)

| Field | Implementation |
|-------|----------------|
| `domain-name.value` | ASCII label rules + **IDNA** (UTS #46) via `idna` |
| `email-addr.value` | **RFC 5322 addr-spec** via `email_address` |
| `url.value` | **WHATWG URL** via `url` crate; `http`/`https`/`ftp` only |

README wire-format table updated accordingly.

---

## Validation profiles (accurate)

| Profile | Checks active |
|---------|----------------|
| `consumer_strict` | All **12** |
| `interop_strict` | All **12** (zero leniency — warnings fail `is_valid()`) |
| `producer_strict` | **11** (skips `References`) |
| `consumer_permissive` | **4** (JSON, type, schema, references) |

This is **intentional profile design**, not a missing implementation.

---

## Release readiness (library scope)

| Milestone | Status after merge |
|-----------|-------------------|
| Core Foundation | Complete (prior work) |
| Data Model + Serialization | Complete (prior work) |
| Pattern Engine (`pattern`) | Complete (prior work) |
| Validation Pipeline (`validate`) | **Complete** (this PR) |
| Graph + Marking + Store | Planned |
| TAXII Client | Planned |

**Distribution:** `rstix` remains in the monorepo; **not on crates.io** (`publish.yml` unchanged).

---

## Known limits

- **SHOULD-level** spec rules emit warnings/info, not errors (maintainer direction #267)
- **`STIX-E0001`** is for JSON syntax/EOF and parse **limit** failures (byte/object/nesting/string caps), not for `ModelError` invariants
- **`from_display_message()`** remains as a legacy fallback for untagged serde messages; primary path is tagged payloads in `serde_error.rs`
- **In-repo conformance corpus** only — not a 1:1 mapping to the external OASIS Interoperability Test document
- **Standalone** leaf deserialize does not capture unknown top-level keys (bundle path does via `x_*` / `common.extra`)

---

## Test plan

- [x] `cargo test -p rstix --features validate --locked`
- [x] `cargo clippy -p rstix --all-targets --features validate --locked -- -D warnings`
- [ ] CI green on PR (workspace matrix)

Key tests:

- `each_diagnostic_code_is_emitted_by_pipeline` — 39/39 codes
- `conformance_valid_corpus_is_valid_under_interop_strict`
- `conformance_invalid_corpus_fails_interop_strict`
- `conformance_versioning_corpus_has_w0003_or_w0004`
- `parse_bridge::model_serde_message_maps_to_e0004_not_e0001`
- `parse_bridge::true_json_syntax_error_maps_to_e0001`
- `conformance_invalid_corpus_never_emits_false_e0001`

---

## Related work

Builds on #265, #268 (Data Model + Serialization), #272 (Pattern Engine), and prior Validation Pipeline check implementation. This PR is the conformance + gap-closure pass only.


crates/rstix/src/model/serde_error.rs
```diff
@@ -0,0 +1,641 @@
...
+/// The human-readable [`Display`](std::fmt::Display) text is preserved at the start so
+/// serde error strings remain readable; the tagged JSON payload follows for round-trip.
+pub(crate) fn encode_for_serde(err: &ModelError) -> String {
+    let payload = serde_json::to_string(&ModelErrorWire::from(err)).expect("model error wire json");
+    format!("{err}{TAG}{payload}")
+}
+
+/// Decode a model error from a serde custom message, if tagged.
+pub(crate) fn decode_from_serde(message: &str) -> Option<ModelError> {
+    let tag_pos = message.find(TAG)?;
```
mostafa: Tag-injection vector. `encode_for_serde` builds `format!("{err}{TAG}{payload}")` — the human-readable `Display` (which interpolates attacker-controlled strings for several variants: `{actual}` in `UnexpectedObjectType`, `{selector}` in `GranularSelectorSyntaxInvalid`, `{detail}` in `ExtensionDeserializeFailed`, `{property}` in the forbidden-property errors) comes *before* the tag, and `decode_from_serde` splits on `message.find(TAG)` — the **first** occurrence.

So a crafted bundle whose field carries `\u{001e}rstix-model\u{001e}{...wire json...}` (control chars are legal in JSON strings via `\u001e`) lands that sequence in the Display prefix. `decode` then finds the injected tag first and recovers a *forged* `ModelError` → forged `DiagnosticCode`/severity. If the forged variant maps to Info/Warning, a genuinely-invalid object's error is downgraded, and under Standard leniency `is_valid()` can wrongly return `true`. For a validator ingesting untrusted CTI that's a diagnostic-forgery / soft validation-bypass.

Minimal fix: `message.rfind(TAG)` — the real payload is always appended last, so the last tag is always the true delimiter and injected content in the Display prefix can't win. Better still, frame payload-first (`format!("{TAG}{payload}{TAG}{err}")` and split on the leading tag) so the boundary never depends on Display content, and/or strip/reject the `\u{001e}` sequence from user-derived fields before they enter a Display. A test with an injected tag in a `selector`/`detail` value would lock it in.

crates/rstix/Cargo.toml
```diff
@@ -23,6 +23,10 @@ thiserror = "2"
 time = { version = "0.3", features = ["formatting", "parsing"] }
 uuid = { version = "1", features = ["v4", "v5"] }
 
+idna = "1.0"
```
mostafa: `idna` / `url` / `email_address` are added as unconditional (non-`optional`) dependencies, and the validators that use them run on the **base deserialize path** (`DomainName`/`EmailAddr`/`Url::validate()` at `sco/*.rs`), not behind the `validate` feature. Two consequences:

1. Footprint: `cargo check -p rstix --no-default-features` now compiles `url` (which itself pulls `idna` + `percent-encoding` + `form_urlencoded`) and `email_address` even though nothing uses them without `serde`. A lean-model consumer (e.g. a `rattack`-style crate on `default-features = false`) pays for them for nothing. This erodes the no-default-features discipline the crate has held across prior PRs.
2. Strictness: with default features (serde on, validate off), a `domain-name`/`email-addr`/`url` SCO whose value isn't full IDNA/WHATWG/RFC-5322 valid now hard-fails base deserialize with `*FormatInvalid`. Value format is a STIX SHOULD, and the parse-permissive / validate-strict split from #267 puts SHOULD checks in the pipeline as warnings — a hard base-deserialize rejection is stricter than that and will drop real-world-but-imperfect CTI for consumers who never opted into validation.

Suggest making the three deps `optional = true` and gating both the deps and the strict format checks behind a feature (the `validate` feature, or a dedicated `wire-format`), with the base path keeping the prior lightweight ASCII checks (or emitting the strict format issue as a SHOULD-warning inside the pipeline rather than a deserialize error). Non-blocking, but worth settling before this is tagged release-ready.

mostafa (commented):
Reviewed the conformance + gap-closure pass in a worktree. Gates green (518 tests, clippy `--all-targets --features validate -D warnings` clean, fmt clean, `cargo check --no-default-features` clean). A lot of this is excellent: the `ModelError` ↔ `ModelErrorWire` mapping is a full 1:1 both directions, `model_mapping.rs` is genuinely exhaustive (no catch-all `_ => E0003`, so a new `ModelError` variant will fail to compile until mapped — exactly right), the per-diagnostic coverage test that fails if `coverage_cases()` drifts from `DiagnosticCode::ALL` is a strong regression guard, and reserving `STIX-E0001` for JSON syntax/limits while routing model invariants to structured codes is the correct split.

Two findings below: one security (the tagged-error channel), one footprint/philosophy (the new wire validators). Neither blocks CI, but the first is worth fixing before merge.

crates/rstix/src/model/serde_error.rs
SecurityEnthusiast: Good catch. agreed this was worth fixing before merge.

**Fix:** Switched to payload-first framing:

`{TAG}{json}{TAG}{display}`

`decode_from_serde` now requires a leading tag and only parses JSON between the first tag pair, so attacker-controlled text in `Display` (selectors, `detail`, etc.) cannot win the delimiter scan.

**Regression test:** `decode_ignores_tag_sequence_embedded_in_display_fields` embeds `{TAG}{forged_wire_json}` inside `ExtensionDeserializeFailed.detail` and asserts decode recovers the genuine error, not the forged variant.

Human-readable text remains in the serde message suffix (existing `rejects_wrong_type_field` tests still pass via `contains(...)`).

crates/rstix/Cargo.toml
SecurityEnthusiast: Implemented the split you suggested.

**Deps:** `idna`, `url`, and `email_address` are now `optional = true` and pulled in only via the `validate` feature:

```toml
validate = ["serde", "pattern", "dep:idna", "dep:url", "dep:email_address"]

SecurityEnthusiast: Worth to mention here as record:

`cargo check -p rstix --no-default-features` no longer compiles those crates.

**Parse vs pipeline:**

| Layer | Behavior |
| --- | --- |
| Parse boundary (`serde`) | Lightweight checks restored (ASCII domain labels, basic `@` email shape, `http`/`https`/`ftp` URL prefix) |
| Validation Pipeline (`validate`) | Strict IDNA / RFC 5322 / WHATWG via `*_format_strict()` in `object_validate` → `STIX-I0002` |

Matches the parse-permissive / validate-strict split from #267.

README wire-format table and CHANGELOG updated.

mostafa (approved):

I fixed the failing test in https://github.com/timescale/rsigma/commit/418178362078f00691415b0d005aef6256ca9d21.

PR #352, opened 2026-07-17
fix(rstix): serde-gate wire-format deps and document DD-DM-001

SecurityEnthusiast:
Make idna, url, email_address, base64, and encoding_rs optional under serde so --no-default-features builds stay lean, while default parse behavior is unchanged. Record wire MUST at parse as DD-DM-001 in crate and library docs.


mostafa (approved):

PR #373, opened 2026-07-21
feat(rstix): TAXII 2.1 client (taxii feature)

SecurityEnthusiast:
## Summary

Adds an async OASIS TAXII 2.1 HTTP client behind the optional **`taxii`** feature on `rstix`, with wiremock integration tests (59) and an optional Docker live harness (6 ignored tests).Workspace tests pass locally.

### TAXII 2.1 client (`taxii` feature)

- **`TaxiiClient`** / **`TaxiiClientConfig`** — all seven normative endpoint groups: Discovery, API Root, Collections, Objects (GET/POST/DELETE + streams), Manifest, Status (202 polling), Versions. **Channels (spec §6, RESERVED) are not implemented.**
- **Wire format** — `TaxiiEnvelope` for object exchange (not `Bundle`); `Accept` / `Content-Type: application/taxii+json;version=2.1`; manifest also accepts STIX JSON; trailing-slash URL rules; HTTPS enforced by default (`allow_insecure_http` for tests); `max_content_length` / `max_response_bytes` limits.
- **Auth** — `BearerAuth`, `BasicAuth`, `ApiKeyHeader` via `TaxiiAuthProvider`; secrets in `secrecy::SecretString` with redacted `Debug`.
- **Pagination** — `TaxiiPaged<T>` with `X-TAXII-Date-Added-First` / `Last`; cursor + header fallback on objects, manifest, object-by-id, and versions streams; HTTP 416 stream recovery; opaque `next` cursors.
- **Resilience** — `RetryPolicy` (503/429/network, `Retry-After`); `PreflightPolicy`; `PostSubmitPolicy` (poll until complete); `CapabilityPolicy` (API Root versions + collection media types); full `TaxiiError` mapping including HTTP 415 and parsed `WWW-Authenticate` on 401.
- **DNS SRV** — `_taxii2._tcp` discovery via `resolve_taxii_srv` / `discover_via_srv` (RFC 2782 weighted selection); `TaxiiClientConfig::dns_nameserver()` for custom resolvers (CoreDNS in live harness).
- **TLS / trust** — single **rustls** stack (TLS 1.2 + 1.3 only): `ServerTrustPolicy` (system roots, SPKI pin, DANE), `TlsaCache`, `build_rustls_config`, clock skew from HTTP `Date` for `added_after`.
- **mTLS** — `ClientCertificate::from_pem` and `from_pkcs12_der` (pure-Rust PKCS#12 via **`p12-keystore`**) embedded in rustls config.
- **rstix** now depends on its own **`reqwest` 0.13** with **`rustls` only**.

### Live integration harness (`crates/rstix/tests/taxii-live/`)

- Docker stack: Wiremock + Caddy + CoreDNS; `run-live-tests.sh` generates certs, starts compose, adds `/etc/hosts` for `dane.taxii.test` / `taxii.test`.
- **6 ignored live tests** in `taxii_live.rs`: TLS discovery, PEM mTLS (localhost SNI — Caddy strict), PKCS#12 mTLS, DNS SRV, TLS 1.3-only listener, DANE (TLSA via CoreDNS template plugin + `dns_nameserver()`).
- Harness fixes: CoreDNS TLSA zone, SRV records, cert file permissions (`chmod 644` on PEM/key/P12 for test runner), zero-config paths in test code, mTLS uses `localhost` not IP for SNI.

### Documentation

- `crates/rstix/README.md` — TAXII API tables, invariant decisions, test commands.
- `CHANGELOG.md` `[Unreleased]` — TAXII client entry.
- `docs/content/library/rstix.md`, `docs/content/reference/feature-flags.md`, `docs/content/reference/architecture.md` — `taxii` feature only; Channels deferral documented.

### Local (run on branch after upstream merge)
- [x] `cargo test -p rstix --features taxii --test taxii_client` (59 passed)
- [x] `cargo clippy -p rstix --all-targets --features taxii -- -D warnings`
- [x] `cargo test --workspace --all-features --locked` (after merge with upstream `main`)
- [x] `cargo check -p rstix --no-default-features --locked` (rstix lean-features CI parity)
- [x] TLS regression: `cargo test -p rstix --features taxii pkcs12` + `cargo test -p rsigma-runtime --all-features -- builds_with_ca_and_identity`
- [ ] `cargo fmt --all -- --check` (CI will enforce; run locally if not yet)
- [ ] `mkdocs build --strict` (docs changed; CI Docs job will enforce)

### Optional (not in CI)
- [x] Live harness: `./crates/rstix/tests/taxii-live/run-live-tests.sh` then `cargo test -p rstix --features taxii --test taxii_live -- --ignored --nocapture` (6 tests: TLS, PEM mTLS, PKCS#12, SRV, TLS 1.3, DANE)

### CI (GitHub Actions on PR)
- Check, MSRV, Format, Clippy, rstix lean features, Test, Coverage, Doc, WASM, Sigma corpus, MCP smoke, Docs site build

## Out of scope / known limits

- TAXII **Channels (§6)** — spec RESERVED; not implemented (documented in README + CHANGELOG).
- Live harness assumes client and Docker on the **same machine** (`127.0.0.1` / `localhost`).


crates/rstix/src/taxii/server_trust.rs
```diff
@@ -0,0 +1,239 @@
...
+                return Err(Error::General("certificate SPKI pin mismatch".into()));
+            }
+            if self.pin_only {
+                return Ok(ServerCertVerified::assertion());
+            }
+        }
+
+        self.inner
+            .verify_server_cert(end_entity, intermediates, server_name, ocsp_response, now)
+    }
```
mostafa: `ServerTrustPolicy::Dane` fails open here: this branch can only accept, never reject.

- TLSA records present but none match `verify_dane` means control falls through to the WebPKI `self.inner.verify_server_cert(...)` below and can still succeed. Per RFC 7671 sections 5-6, when usable TLSA records exist and none match, the client MUST reject. As written, an attacker holding any publicly-CA-issued cert for the host defeats the TLSA pin.
- No TLSA records for the host (the cache is populated only by caller code via `TlsaCache::insert`, so this verifier is the sole enforcement point) means a silent downgrade to PKIX with no error, even though the caller explicitly selected `Dane`.

Net effect: `Dane` provides no security over `SystemRoots` while signaling that it does. It should fail closed: when usable records exist and none match, return `Err`; when none are present, either error or require the caller to explicitly opt into a PKIX fallback.

crates/rstix/src/taxii/dane.rs
```diff
@@ -0,0 +1,79 @@
...
+    Some(Sha256::digest(&spki).into())
+}
+
+/// Verify `end_entity` against DNSSEC-validated TLSA records.
+pub fn verify_dane(end_entity: &CertificateDer<'_>, records: &[TlsaRecord]) -> bool {
+    records
+        .iter()
+        .filter(|r| r.acceptable_for_tls_client())
+        .any(|r| r.matches_certificate(end_entity))
+}
```
mostafa: DANE usage-type semantics are collapsed to DANE-EE. `verify_dane` accepts usages 0-3 uniformly and, on any match, the caller bypasses PKIX entirely, but `matches_certificate` above always compares against `end_entity`:

- Usage 1 (PKIX-EE) is accepted without the PKIX chain/name/expiry validation that usage mandates.
- Usages 0/2 (PKIX-TA / DANE-TA) associate a trust anchor, not the end-entity, so comparing them against `end_entity` essentially never matches and those deployments silently fall through to PKIX.

Only usage 3 (DANE-EE) is handled correctly. Please distinguish the usages (DANE-EE bypasses PKIX; PKIX-EE requires a match and PKIX; TA usages require chain-to-anchor plus name checks), or support DANE-EE only and reject other usages rather than mis-handling them.

crates/rstix/src/taxii/server_trust.rs
```diff
@@ -0,0 +1,239 @@
...
+        if !self.pins.is_empty() {
+            let Some(hash) = spki_sha256(end_entity) else {
+                return Err(Error::General("failed to hash SPKI".into()));
+            };
+            if !self.pins.contains(&hash) {
+                return Err(Error::General("certificate SPKI pin mismatch".into()));
+            }
+            if self.pin_only {
+                return Ok(ServerCertVerified::assertion());
+            }
```
mostafa, line 221: Minor: this `pin_only` path (and the DANE-EE accept at line 165) returns `ServerCertVerified::assertion()` with no hostname (SAN) or expiry check. For DANE-EE that is spec-permitted (RFC 7671 section 5.1), and for pin-only it is arguably intended, but it is worth documenting since a pinned-but-expired or wrong-host certificate would be accepted.

crates/rstix/src/taxii/request.rs
```diff
@@ -0,0 +1,260 @@
...
+        }
+        let clock_skew = headers
+            .get(reqwest::header::DATE)
+            .and_then(|v| v.to_str().ok())
+            .and_then(parse_http_date_skew);
+        if let (Some(skew), Ok(mut guard)) = (clock_skew, self.clock_skew.write()) {
+            *guard = Some(skew);
+        }
+        let bytes = response.bytes().await.map_err(TaxiiError::NetworkError)?;
+        if bytes.len() > self.max_response_bytes {
```
mostafa: Minor (DoS): the size cap is checked after the body is fully buffered (`bytes.len()`), so a hostile server can force the entire response into memory before this rejects it. Consider checking `Content-Length` up front or capping mid-stream against `max_response_bytes`.

mostafa (commented):
Reviewed the TAXII 2.1 client. Overall a strong PR: clean feature gating (`taxii` fully optional), `reqwest` pinned to rustls-only and scoped to rstix, HTTPS enforced by default including on server-provided URLs, TLS 1.2+ only, secrets in `SecretString` with redacted Debug, no `danger_accept_invalid_certs` anywhere, pure-Rust PKCS#12, and 59 wiremock tests plus clippy and fmt all green.

One real issue to resolve before merge: the `ServerTrustPolicy::Dane` trust mode is unsound (it fails open and collapses the TLSA usage types to DANE-EE). Details inline. Since DANE is opt-in and the default `SystemRoots` and SPKI-pinning paths are fine, the rest looks good to me. Two minor inline notes as well.

crates/rstix/src/taxii/server_trust.rs
SecurityEnthusiast: Fail-closed end-to-end. No PKIX fallback when `Dane` is selected.
| Situation | Behavior |
|-----------|----------|
| TLSA DNS lookup returns no records | `TaxiiError::InvalidServerTrust` in `prefetch_dane` (`request.rs`) |
| TLSA cache empty at handshake | `Error::General` — no WebPKI fallback |
| Usable TLSA records, none match | `evaluate_dane` returns `Err` — handshake fails |
| Usage 0 or 1 association matches | `RequirePkix` → `inner.verify_server_cert` only on that path |

No explicit “DANE with PKIX fallback” mode was added. Tests in `dane.rs` cover fail-closed on mismatch.

crates/rstix/src/taxii/dane.rs
SecurityEnthusiast: Full RFC 7671 usage 0–3 (not DANE-EE-only).
| Usage | Name | Behavior |
|-------|------|----------|
| 3 | DANE-EE | EE association match → accept without PKIX |
| 2 | DANE-TA | Matching chain cert as trust anchor → chain verified → accept without PKIX |
| 1 | PKIX-EE | EE match → require PKIX (`inner.verify_server_cert`) |
| 0 | PKIX-TA | Chain cert match → require PKIX |

Association matching uses `TlsaRecord::association_matches` (selector 0/1, matching 0/1). Usages 0 and 2 match against the presented chain, not only the EE cert.
Tests: DANE-EE → accept; PKIX-EE → require PKIX; mismatch → fail-closed.

crates/rstix/src/taxii/request.rs
SecurityEnthusiast: Both defenses in `read_response_body`:

1. **`check_content_length`** — reject when `Content-Length` exceeds `max_response_bytes` before reading.

2. **Streaming cap** — `bytes_stream()` with per-chunk limit (covers missing/wrong `Content-Length` and chunked bodies).

Unit test: `content_length_over_max_is_rejected_before_read`. Wiremock test `rejects_response_larger_than_max_bytes` covers the streaming path.

crates/rstix/src/taxii/server_trust.rs
SecurityEnthusiast: Documented, not changed — behavior is intentional.

- **`PinnedSpkiOnly`:** SPKI pin is the trust decision (TAXII 2.1 spec section 8.5.2 certificate pinning).
- **DANE-EE / verified DANE-TA:** RFC 7671 section 5.1 permits skipping PKIX name/expiry on those paths.

Documentation added in:

- `server_trust.rs` module and enum rustdoc
- Inline comments on `AcceptWithoutPkix` and `pin_only` branches
- README invariant table

SAN/expiry enforcement on these paths would contradict RFC 7671 and break intentional pin-only use (including live harness SPKI pins).

mostafa (approved):

PR #387, opened 2026-07-24
feat(rstix): TAXII collection ingest into StixStore (taxii-store)

SecurityEnthusiast:
## Summary

- Add **`taxii-store`** meta-feature (`taxii` + `store`) with `ingest_collection` / `ingest_collection_with_bundle_id` — paginated `objects_stream` → synthetic `Bundle` → `StixStore::import_bundle`.
- TAXII capability: accept `application/taxii+json` collection media types; default absent `media_types` to `application/stix+json` (§5.2.1); enforce API Root `max_content_length` > 0.
- Store: `import_bundle` resolves refs against objects already in the store (incremental ingest); Wiremock tests for ingest + store ref behavior.

## Test plan

- [x] `RUSTDOCFLAGS="-D warnings -D rustdoc::broken-intra-doc-links" cargo doc --workspace --all-features --locked --no-deps`
- [x] `RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- [x] `RUSTFLAGS=-Dwarnings cargo test --workspace --all-features --locked`
- [x] `RUSTFLAGS=-Dwarnings cargo test -p rstix --features taxii-store --test taxii_store --locked`
- [x] `npm ci && npm run docs:build && npm run docs:validate`

### DIFF
--- a/crates/rstix/src/taxii/ingest.rs
+++ b/crates/rstix/src/taxii/ingest.rs
@@ -2,12 +2,12 @@
 //!
 //! Requires **`taxii`** and **`store`** features (or the `taxii-store` meta-feature).
 
-use futures::StreamExt;
-
 use crate::core::StixId;
-use crate::model::Bundle;
-use crate::store::{ImportReport, StixStore, StoreError};
+use crate::store::{ImportReport, StixStore, StoreError, audit_unresolved_refs};
 
+use super::envelope::TaxiiEnvelope;
+use super::headers::TaxiiPageHeaders;
+use super::pagination::{advance_more_page, recover_from_range_not_satisfiable};
 use super::{TaxiiClient, TaxiiError, TaxiiFilter};
 
 /// Synthetic bundle id for [`ingest_collection`] when no custom id is supplied.
@@ -26,12 +26,12 @@ pub enum IngestError {
 
 /// Fetch all objects from a TAXII collection (paginated) and import them into `store`.
 ///
-/// Objects are collected from [`TaxiiClient::objects_stream`], wrapped in a synthetic
-/// [`Bundle`] via [`Bundle::from_objects`], then passed to [`StixStore::import_bundle`].
-/// Cross-object references are checked against the fetched set and objects already
-/// in the store (incremental TAXII sync).
+/// Each TAXII response page is upserted via [`StixStore::import_objects`] so memory stays
+/// bounded by page size. After the full collection is imported, references are audited once
+/// against the store.
 ///
-/// Uses [`DEFAULT_INGEST_BUNDLE_ID`] as the wrapper bundle id. For a custom id, call
+/// Uses [`DEFAULT_INGEST_BUNDLE_ID`] as the conventional wrapper id when exporting the store
+/// with [`StixStore::export_bundle`]. For a custom id, call
 /// [`ingest_collection_with_bundle_id`].
 pub async fn ingest_collection(
     client: &TaxiiClient,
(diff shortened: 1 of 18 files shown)


mostafa (commented):
Reviewed the TAXII collection ingest. Most of this is clean and mergeable: the `taxii-store` gating is correct, the store-aware `import_bundle` ref resolution is exactly right for incremental sync, the `store-fs` gating of `StoreError::io` removes a latent dead-code warning, and the capability media-type handling is spec-correct (§5.2.1 default, taxii+json and stix+json accepted, `max_content_length > 0`). Gates green: 455 tests pass, clippy `--all-features -- -D warnings` clean, fmt clean.

One memory-bounds concern on the ingest path, inline. Not a correctness bug, but worth resolving before this is the recommended ingest path for ATT&CK-scale feeds.

crates/rstix/src/taxii/ingest.rs
```diff
@@ -0,0 +1,70 @@
...
+    filter: TaxiiFilter,
+    bundle_id: StixId,
+) -> Result<ImportReport, IngestError> {
+    let mut objects = Vec::new();
+    let mut stream = client.objects_stream(api_root_url, collection_id, filter);
+    while let Some(result) = stream.next().await {
+        objects.push(result?);
+    }
+    let bundle = Bundle::from_objects(bundle_id, objects);
+    Ok(store.import_bundle(&bundle)?)
```
mostafa: This drains the entire paginated stream into `objects` and builds a single synthetic `Bundle` holding all of it before importing. `objects_stream` deliberately pages to keep memory flat, so buffering the whole collection here undoes that: for a large collection (full MITRE ATT&CK, or a high-volume threat feed) this is unbounded client-side memory, and there is no cap on object count, so a hostile or just large TAXII server can OOM the ingesting client. Same class of issue as the response-body buffering from #373, but now spanning the whole collection.

Since the store already resolves refs incrementally against itself (`ref_target_resolved`), the natural fix is to import in batches (per page, or per N objects) rather than accumulating everything. The one tradeoff: a forward reference from an early page to an object on a later page would then be reported as unresolved, but those are advisory-only and the objects still land in the store, so batching is safe. If you would rather preserve whole-collection ref reporting, at minimum add a configurable max-object cap that errors out (mirroring `max_response_bytes`) so this buffer cannot grow without bound.

SecurityEnthusiast: Thanks for the review. The unbounded-memory concern is addressed.

### Problem

The original implementation drained the full paginated TAXII stream into a `Vec` before calling `import_bundle`, so memory grew with collection size rather than page size.

Per-page `import_bundle` alone was not enough: ref audit ran **per page**, so forward references (e.g. a report on page 1 referencing an indicator on page 2) were reported as unresolved even after a successful ingest.

### Fix

Ingest now uses a two-phase pattern:

1. **Per page:** upsert via `StixStore::import_objects` — memory stays bounded by page size; no synthetic `Bundle` per page.
2. **After all pages:** run a single ref audit via shared `audit_unresolved_refs(store, ingested_ids)` against the full store.

`import_bundle` was refactored to the same model: upsert first (`import_objects`), then audit once (`audit_unresolved_refs`). Both bundle import and TAXII ingest share one ref-audit code path.

### Result

- Memory is bounded by TAXII page size, not collection size.
- Forward refs across pages resolve correctly in the final `ImportReport`.
- No false positives from premature per-page ref checks.

### Test coverage

Added `ingest_collection_resolves_forward_refs_across_pages`: page 1 is a report referencing an indicator; page 2 contains the indicator. Asserts `unresolved_references` is empty after full ingest.

Existing paginated ingest and idempotency tests still pass.

mostafa (approved):

PR #399, opened 2026-07-26
✨ feat: support list values in add_condition pipeline transformation

frack113:
## Summary

Implement list value support in the `add_condition` pipeline transformation, matching pySigma's `AddConditionTransformation` behavior. This allows pipeline definitions to specify multiple OR-linked values per field (e.g., `EventID: [12, 13, 14]`) instead of requiring separate transformations for each value.

### 🐛 Parsing & Semantics

- `parse_value_mapping` now accepts YAML sequences as values, producing `HashMap<String, Vec<SigmaValue>>` instead of `HashMap<String, SigmaValue>`
- Extracted `yaml_to_sigma_value` helper for converting a single YAML value to `SigmaValue`
- Added guard against empty sequences: `[]` produces no entry rather than a silent no-match detection item

### 📋 Sysmon Pipeline Alignment

Aligned the sysmon pipeline with the [Sigma taxonomy](https://github.com/SigmaHQ/sigma-specification/blob/main/specification/sigma-appendix-taxonomy.md#windows-folder):

| Category | EventIDs | Status |
|----------|----------|--------|
| `sysmon_status` | 4, 16 | **Added** |
| `registry_event` | 12, 13, 14 | **Added** (previously only separate sub-categories) |
| `wmi_event` | 19, 20, 21 | **Added** |
| `pipe_created` | 17, **18** | Updated — missing EventID 18 |
| `file_block_shredding` | 28 | **Added** |

### ✅ Tests

- Unit test: `parse_add_condition_with_list_values` — validates YAML sequence parsing
- Unit test: `parse_add_condition_with_empty_sequence` — validates empty sequence handling
- E2E test: `test_pipeline_add_condition_list_values_e2e` — full engine pipeline test with `registry_event` and `EventID: [12, 13, 14]`

## Commits

```
0511cef2 📋 fix(sysmon): align pipeline with Sigma taxonomy
1f1cfa49 ✅ test: add e2e test for add_condition with list values
c9ac2b26 🐛 fix: guard empty YAML sequences in parse_value_mapping
1df9d3ac ✨ feat: support list values in add_condition pipeline transformation
```

## Verification

```
696 passed, 0 failed
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,23 +6,11 @@ All notable changes to RSigma are documented in this file. Each entry correspond
 
 ### Support list values in `add_condition` pipeline transformation (#399)
 
-- `add_condition` now accepts YAML sequences for field values:
-  `conditions: {EventID: [12, 13, 14]}` instead of only `conditions: {EventID: 12}`.
-  Values within a field are OR-linked, matching pySigma's `AddConditionTransformation`.
-- **Breaking**: `AddCondition.conditions` changed from `HashMap<String, SigmaValue>` to
-  `HashMap<String, Vec<SigmaValue>>`. Downstream constructors using the `Transformation` enum
-  (re-exported as `rsigma_eval::pipeline::Transformation`) must update their field value types.
-- **Sysmon pipeline additions** (aligned with [Sigma taxonomy](https://github.com/SigmaHQ/sigma-specification/blob/main/specification/sigma-appendix-taxonomy.md#windows-folder)):
-  - `sysmon_status` (EventID 4, 16), `registry_event` (12, 13, 14), `wmi_event` (19, 20, 21),
-    `file_block_shredding` (28) — new categories added.
-  - `pipe_created` updated to include EventID 18 (missing).
-  - **Behavior change**: rules with `category: registry_event`, `wmi_event`, or `sysmon_status`
-    previously got no `EventID` condition at all and matched every Sysmon event; they now match
-    only the relevant EventIDs. Rules with `category: pipe_created` now also match EventID 18.
-- `[]` (empty sequence) is now a parse error instead of silently skipping, preventing pipeline
-  typos from widening rule scope.
-- New tests: `parse_add_condition_with_list_values`, `parse_add_condition_with_empty_sequence`,
-  `test_pipeline_add_condition_list_values_e2e`.
+- `add_condition` accepts YAML sequences for field values (`conditions: {EventID: [12, 13, 14]}`), and the values of one field are OR-linked, matching pySigma's `AddConditionTransformation`. Pipelines no longer need one transformation per value.
+- An empty sequence (`field: []`) and a non-scalar value (a mapping, or a sequence nested inside a sequence) are load-time errors naming the offending field. Both previously produced a condition that silently matched nothing or disappeared, which widened the rule instead of failing.
+- **Breaking**: `AddCondition.conditions` changed from `HashMap<String, SigmaValue>` to `HashMap<String, Vec<SigmaValue>>`. Downstream code constructing the `Transformation` enum (re-exported as `rsigma_eval::pipeline::Transformation`) must wrap scalar values in a `Vec`.
+- The builtin `sysmon` pipeline gains the [Sigma taxonomy](https://github.com/SigmaHQ/sigma-specification/blob/main/specification/sigma-appendix-taxonomy.md#windows-folder) categories it was missing: `sysmon_status` (EventID 4, 16), `registry_event` (12, 13, 14), `wmi_event` (19, 20, 21), `file_block_shredding` (28), and `sysmon_error` (255). `pipe_created` now also covers EventID 18.
+- **Behavior change**: rules carrying one of those five categories previously got no `EventID` condition at all and matched every Sysmon event; they now match only the relevant EventIDs. Rules with `category: pipe_created` also match EventID 18.
 
 ### OCSF Detection Finding output (#397)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,26 @@ All notable changes to RSigma are documented in this file. Each entry correspond
 
 ## [Unreleased]
 
+### Support list values in `add_condition` pipeline transformation (#399)
+
+- `add_condition` now accepts YAML sequences for field values:
+  `conditions: {EventID: [12, 13, 14]}` instead of only `conditions: {EventID: 12}`.
+  Values within a field are OR-linked, matching pySigma's `AddConditionTransformation`.
+- **Breaking**: `AddCondition.conditions` changed from `HashMap<String, SigmaValue>` to
+  `HashMap<String, Vec<SigmaValue>>`. Downstream constructors using the `Transformation` enum
+  (re-exported as `rsigma_eval::pipeline::Transformation`) must update their field value types.
+- **Sysmon pipeline additions** (aligned with [Sigma taxonomy](https://github.com/SigmaHQ/sigma-specification/blob/main/specification/sigma-appendix-taxonomy.md#windows-folder)):
+  - `sysmon_status` (EventID 4, 16), `registry_event` (12, 13, 14), `wmi_event` (19, 20, 21),
+    `file_block_shredding` (28) — new categories added.
+  - `pipe_created` updated to include EventID 18 (missing).
+  - **Behavior change**: rules with `category: registry_event`, `wmi_event`, or `sysmon_status`
+    previously got no `EventID` condition at all and matched every Sysmon event; they now match
+    only the relevant EventIDs. Rules with `category: pipe_created` now also match EventID 18.
+- `[]` (empty sequence) is now a parse error instead of silently skipping, preventing pipeline
+  typos from widening rule scope.
+- New tests: `parse_add_condition_with_list_values`, `parse_add_condition_with_empty_sequence`,
+  `test_pipeline_add_condition_list_values_e2e`.
+
 ### OCSF Detection Finding output (#397)
 
 - Any line-oriented sink can emit [OCSF](https://ocsf.io) Detection Finding (class 2004) JSON instead of native NDJSON: `--output 'file:///findings.ocsf.ndjson?format=ocsf'`, or the same suffix on a `daemon.output.sinks` entry. Findings then flow natively into Splunk, Elastic, CrowdStrike NG-SIEM, and anything else that consumes OCSF.
--- a/crates/rsigma-eval/README.md
+++ b/crates/rsigma-eval/README.md
@@ -369,7 +369,7 @@ Each transformation item in a pipeline can have:
 | `field_name_suffix` | `suffix` | Add a suffix to all field names |
 | `field_name_transform` | `transform_func`, `mapping` | Case transformation (see below) |
 | `drop_detection_item` | — | Remove matching detection items |
-| `add_condition` | `conditions: {k: v | [v1, v2, ...]}`, `negated` (default: `false`) | Inject additional detection conditions; list values are OR-linked |
+| `add_condition` | `conditions: {k: v \| [v1, v2, ...]}`, `negated` (default: `false`) | Inject additional detection conditions; list values are OR-linked |
 | `change_logsource` | `category`, `product`, `service` | Modify logsource fields |
 | `replace_string` | `regex`, `replacement`, `skip_special` (default: `false`) | Regex-based string replacement (`skip_special` preserves wildcards) |
 | `map_string` | `mapping: {k: v \| [v1, v2]}` | Map string values to replacements (supports one-to-many) |
--- a/crates/rsigma-eval/README.md
+++ b/crates/rsigma-eval/README.md
@@ -369,7 +369,7 @@ Each transformation item in a pipeline can have:
 | `field_name_suffix` | `suffix` | Add a suffix to all field names |
 | `field_name_transform` | `transform_func`, `mapping` | Case transformation (see below) |
 | `drop_detection_item` | — | Remove matching detection items |
-| `add_condition` | `conditions: {k: v}`, `negated` (default: `false`) | Inject additional detection conditions |
+| `add_condition` | `conditions: {k: v | [v1, v2, ...]}`, `negated` (default: `false`) | Inject additional detection conditions; list values are OR-linked |
 | `change_logsource` | `category`, `product`, `service` | Modify logsource fields |
 | `replace_string` | `regex`, `replacement`, `skip_special` (default: `false`) | Regex-based string replacement (`skip_special` preserves wildcards) |
 | `map_string` | `mapping: {k: v \| [v1, v2]}` | Map string values to replacements (supports one-to-many) |
--- a/crates/rsigma-eval/pipelines/sysmon.yml
+++ b/crates/rsigma-eval/pipelines/sysmon.yml
@@ -46,6 +46,17 @@ transformations:
         category: network_connection
         product: windows
 
+  - id: sysmon_sysmon_status
+    type: add_condition
+    conditions:
+      EventID:
+        - 4
+        - 16
+    rule_conditions:
+      - type: logsource
+        category: sysmon_status
+        product: windows
+
   - id: sysmon_process_termination
     type: add_condition
     conditions:
@@ -109,6 +120,18 @@ transformations:
         category: file_event
         product: windows
 
+  - id: sysmon_registry_event
+    type: add_condition
+    conditions:
+      EventID:
+        - 12
+        - 13
+        - 14
+    rule_conditions:
+      - type: logsource
+        category: registry_event
+        product: windows
+
   - id: sysmon_registry_add
     type: add_condition
     conditions:
@@ -157,10 +180,24 @@ transformations:
   - id: sysmon_pipe_created
     type: add_condition
     conditions:
-      EventID: 17
+      EventID:
+        - 17
+        - 18
     rule_conditions:
       - type: logsource
         category: pipe_created
+        product: windows
+
+  - id: sysmon_wmi_event
+    type: add_condition
+    conditions:
+      EventID:
+        - 19
+        - 20
+        - 21
+    rule_conditions:
+      - type: logsource
+        category: wmi_event
         product: windows
 
   - id: sysmon_dns_query
@@ -217,6 +254,15 @@ transformations:
         category: file_block_executable
         product: windows
 
+  - id: sysmon_file_block_shredding
+    type: add_condition
+    conditions:
+      EventID: 28
+    rule_conditions:
+      - type: logsource
+        category: file_block_shredding
+        product: windows
+
   - id: sysmon_file_executable_detected
     type: add_condition
     conditions:
@@ -224,6 +270,15 @@ transformations:
     rule_conditions:
       - type: logsource
         category: file_executable_detected
+        product: windows
+
+  - id: sysmon_sysmon_error
+    type: add_condition
+    conditions:
+      EventID: 255
+    rule_conditions:
+      - type: logsource
+        category: sysmon_error
         product: windows
 
   # Change logsource to sysmon service for all matched categories
--- a/crates/rsigma-eval/src/pipeline/builtin.rs
+++ b/crates/rsigma-eval/src/pipeline/builtin.rs
@@ -69,4 +69,96 @@ mod tests {
         assert!(names.contains(&"fibratus_windows"));
         assert!(names.contains(&"sysmon"));
     }
+
+    fn sysmon_eventid_route(category: &str) -> Option<Vec<i64>> {
+        use crate::pipeline::Transformation;
+        use rsigma_parser::SigmaValue;
+
+        let pipeline = resolve_builtin("sysmon").unwrap().unwrap();
+        for item in &pipeline.transformations {
+            if let Transformation::AddCondition { conditions, .. } = &item.transformation {
+                if let Some(evts) = conditions.get("EventID") {
+                    if let Some(rule_cond) = item.rule_conditions.first() {
+                        if let crate::pipeline::RuleCondition::Logsource {
+                            category: Some(cat),
+                            ..
+                        } = &rule_cond.condition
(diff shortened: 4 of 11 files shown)


crates/rsigma-eval/src/pipeline/parsing.rs
```diff
@@ -791,33 +791,48 @@ fn parse_string_or_list_mapping(
     Ok(map)
 }
 
-fn parse_value_mapping(value: Option<&yaml_serde::Value>) -> Result<HashMap<String, SigmaValue>> {
+fn parse_value_mapping(value: Option<&yaml_serde::Value>) -> Result<HashMap<String, Vec<SigmaValue>>> {
```
mostafa: `cargo fmt` wants this wrapped, which is what the Format job is failing on.

```suggestion
fn parse_value_mapping(
    value: Option<&yaml_serde::Value>,
) -> Result<HashMap<String, Vec<SigmaValue>>> {
```

crates/rsigma-eval/src/pipeline/parsing.rs
```diff
@@ -791,33 +791,48 @@ fn parse_string_or_list_mapping(
...
+                        .iter()
+                        .map(yaml_to_sigma_value)
+                        .collect(),
+                    // Scalar → single-element Vec
+                    other => vec![yaml_to_sigma_value(other)],
                 };
-                map.insert(key.to_string(), sv);
+                if !values.is_empty() {
+                    map.insert(key.to_string(), values);
+                }
```
mostafa: Dropping the key means the condition silently disappears, which makes the rule wider rather than narrower. I ran this on both branches with

```yaml
conditions:
  EventID: []
  Channel: Security
```

and a `registry_event` rule matches an event carrying `EventID: 9999` on this branch, while on `main` the same pipeline matches nothing (the old `SigmaValue::Null` never matched, so it failed closed).

When the empty sequence is the only condition it is not caught here either: it surfaces later as `InvalidModifiers("AllOf detection must not be empty (vacuous truth)")` from the compiler, which points the operator at the compiler rather than at the offending pipeline line.

I would return a parse error naming the field instead, so a pipeline with an empty sequence is rejected at load time.

crates/rsigma-eval/src/pipeline/parsing.rs
```diff
@@ -791,33 +791,48 @@ fn parse_string_or_list_mapping(
...
+                SigmaValue::Float(f)
+            } else {
+                SigmaValue::Null
+            }
+        }
+        yaml_serde::Value::Bool(b) => SigmaValue::Bool(*b),
+        yaml_serde::Value::Null => SigmaValue::Null,
+        _ => SigmaValue::Null,
+    }
+}
```
mostafa: This duplicates the public `SigmaValue::from_yaml` in `crates/rsigma-parser/src/value.rs`. The two are identical except the fallback arm: `from_yaml` turns an unsupported node into a debug-formatted string, this one into `Null`. Worth reusing it, or keeping the local helper with a comment explaining why the fallback has to differ, because as written the divergence reads as accidental.

crates/rsigma-eval/src/pipeline/transformations/mod.rs
```diff
@@ -55,8 +55,14 @@ pub enum Transformation {
...
     /// Add field=value conditions to the rule's detection.
+    ///
+    /// Each value is a `Vec<SigmaValue>` to support list values (OR semantics).
+    /// A single-element vec behaves identically to the old `SigmaValue` scalar.
+    /// A multi-element vec creates a detection item with multiple values, which
+    /// are OR-linked per Sigma semantics — matching pySigma's
+    /// `AddConditionTransformation` behavior.
     AddCondition {
-        conditions: HashMap<String, SigmaValue>,
+        conditions: HashMap<String, Vec<SigmaValue>>,
```
mostafa, line 65: `Transformation` is re-exported as `rsigma_eval::pipeline::Transformation`, so this is a breaking change to a published public API: anything downstream that constructs `AddCondition` with a scalar stops compiling. Please add a `CHANGELOG.md` entry under `## [Unreleased]` that calls that out explicitly, alongside the behavior change to the sysmon pipeline.

crates/rsigma-eval/pipelines/sysmon.yml
```diff
@@ -154,10 +177,24 @@ transformations:
...
+    type: add_condition
+    conditions:
+      EventID:
+        - 19
+        - 20
+        - 21
+    rule_conditions:
+      - type: logsource
+        category: wmi_event
+        product: windows
```
mostafa: Three small things on the pipeline changes.

The file is otherwise ordered by EventID, and this block (19-21) sits between `create_stream_hash` (15) and `pipe_created` (17).

While aligning with the taxonomy, `sysmon_error: 255` is the one category still missing from the pipeline.

Nothing exercises the builtin sysmon pipeline's routing: `builtin.rs` only asserts that it parses, and the new e2e test defines its own inline pipeline. A test that resolves the `sysmon` builtin and asserts the new categories route to the right EventIDs would lock these mappings in.

crates/rsigma-eval/src/pipeline/transformations/tests.rs
```diff
@@ -1176,7 +1176,7 @@ fn test_add_condition_negated() {
     let mut conds = HashMap::new();
     conds.insert(
         "User".to_string(),
-        SigmaValue::String(SigmaString::new("SYSTEM")),
+        vec![SigmaValue::String(SigmaString::new("SYSTEM"))],
```
mostafa, line 1179: `negated` now negates an OR group rather than a single value, so `NOT (a OR b)` is the semantics most likely to regress later. A companion case here with a multi-value list would be worth adding.

crates/rsigma-eval/src/pipeline/transformations/tests.rs
```diff
@@ -1932,3 +1932,60 @@ fn test_case_transformation_snake_case_already_lowercase() {
...
+        conditions, ..
+    } = &pipeline.transformations[0].transformation
+    else {
+        panic!("expected AddCondition");
+    };
+    // Empty sequence should not appear in the conditions map
+    assert!(!conditions.contains_key("field_a"));
+    // Non-empty entry should still be present
+    assert!(conditions.contains_key("field_b"));
+}
```
mostafa, line 2045: If the parser rejects empty sequences (see the comment on `parse_value_mapping`), this test flips into asserting that `parse_pipeline` returns an error and that the message names `field_a`.

mostafa (changes requested):
Good catch on the taxonomy gaps, and the core semantics are right: `add_conditions` puts every value on a single `DetectionItem` and the compiler ORs multiple values by default (`optimize_any_of`), so `EventID: [12, 13, 14]` behaves like pySigma's `AddConditionTransformation`. I checked all five mapping changes against the Sigma taxonomy appendix and the pySigma sysmon pipeline, and they are correct.

Three things before this can merge.

**`cargo fmt --check` fails**, which is the red Format job. Three spots (the signature, the `Sequence` match arm, and both new `let ... else` destructures). `cargo fmt --all` fixes all of them.

**The empty-sequence guard widens rules instead of rejecting them.** Details inline, but the short version: with `conditions: {EventID: [], Channel: Security}` a rule now matches an event carrying `EventID: 9999`, where on `main` the same pipeline matched nothing. A pipeline typo turning into "route everything" is the wrong failure direction for a detection engine.

**Missing `CHANGELOG.md` entry**, and it needs to note two things beyond the feature itself. `Transformation` is re-exported as `rsigma_eval::pipeline::Transformation`, so changing `AddCondition.conditions` to `HashMap<String, Vec<SigmaValue>>` is a breaking change for downstream constructors. And the sysmon additions change matching for existing rules rather than only adding coverage: `registry_event`, `wmi_event`, and `sysmon_status` rules previously got no `EventID` condition at all and matched every Sysmon event, and `pipe_created` rules now also match EventID 18.

Docs are out of sync in three places, and CI gates on the docs build:

- `docs/content/reference/builtin-pipelines.md` still lists `pipe_created | EventID: 17` and is missing the four new categories.
- `crates/rsigma-eval/README.md` documents `add_condition` as `conditions: {k: v}`; the list form should be shown the way the `map_string` row already does.
- `docs/content/guide/processing-pipelines.md` describes `add_condition` as injecting `EventID: 1` and can mention the list form.

Everything else is green locally: clippy clean, tests pass, corpus regression passes.

crates/rsigma-eval/src/pipeline/parsing.rs
frack113: The empty-sequence guard in parse_value_mapping now rejects it with a clear message naming the field, and the parse_add_condition_with_empty_sequence test is flipped to assert the error instead of silently skipping.

crates/rsigma-eval/src/pipeline/parsing.rs
frack113: Good point about the duplication. I removed yaml_to_sigma_value entirely and replaced it with SigmaValue::from_yaml from rsigma-parser. The only difference in fallback behavior (SigmaValue::Null vs SigmaString::new(&format!("{v:?}"))) is fine here since the pipeline parsing layer should preserve non-standard yaml nodes as strings — which from_yaml already does. Also removed the now-unused SigmaString import.

crates/rsigma-eval/src/pipeline/transformations/mod.rs
frack113: the CHANGELOG entry under [Unreleased] explicitly calls out the breaking change to AddCondition.conditions (HashMap<String, SigmaValue> → HashMap<String, Vec<SigmaValue>>) alongside the sysmon behavior changes, as you requested.

crates/rsigma-eval/pipelines/sysmon.yml
frack113: Good catch on the ordering and the missing sysmon_error category. I moved sysmon_wmi_event to after pipe_created (17-18 → 19-21 → 22) to restore the EventID ordering, and added the sysmon_error (255) routing at the end. I also added two routing tests in builtin.rs (sysmon_builtin_routes_new_categories and sysmon_builtin_eventid_order) that exercise the builtin sysmon pipeline's routing and lock these mappings in.

crates/rsigma-eval/src/pipeline/transformations/tests.rs
frack113: Good call — the existing negated test only uses a single value. I added a companion test with a multi-value list (EventID: [1, 2, 3] + negated: true) that verifies the AND NOT (<named detection>) structure and asserts the named detection carries all 3 integer values.

crates/rsigma-eval/src/pipeline/transformations/tests.rs
frack113: Good catch — the test now asserts that the error message names field_a (not just "empty sequence"), matching the parse_value_mapping error format: add_condition: empty sequence for field '{field_a}' is not allowed....

crates/rsigma-eval/src/pipeline/builtin.rs
```diff
@@ -69,4 +69,96 @@ mod tests {
...
+                                            SigmaValue::Integer(n) => Some(*n),
+                                            _ => None,
+                                        })
+                                        .collect(),
+                                );
+                            }
+                        }
+                    }
+                }
+            }
```
mostafa: This is what turned the Clippy job red: `collapsible_if` fires five times on this nested `if let` chain.

A let-chain collapses the whole thing and matches what the crate already does elsewhere:

```rust
if let Transformation::AddCondition { conditions, .. } = &item.transformation
    && let Some(evts) = conditions.get("EventID")
    && let Some(rule_cond) = item.rule_conditions.first()
    && let crate::pipeline::RuleCondition::Logsource { category: Some(cat), .. } =
        &rule_cond.condition
    && cat == category
{
    return Some(/* ... */);
}
```

crates/rsigma-eval/src/pipeline/builtin.rs
```diff
@@ -69,4 +69,96 @@ mod tests {
...
+                None
+            })
+            .collect();
+
+        for i in 0..add_conditions.len().saturating_sub(1) {
+            let (_, prev_nums) = &add_conditions[i];
+            let (_, next_nums) = &add_conditions[i + 1];
+            let prev_min = prev_nums.iter().copied().min().unwrap();
+            let next_min = next_nums.iter().copied().min().unwrap();
+            assert!(
```
mostafa, line 141: Separate from the CI failure, and not urgent: clippy on the 1.88 MSRV toolchain also flags `uninlined_format_args` on this `assert!`. Inlining the arguments into the format string keeps both toolchains quiet.

crates/rsigma-eval/README.md
```diff
@@ -369,7 +369,7 @@ Each transformation item in a pipeline can have:
 | `field_name_suffix` | `suffix` | Add a suffix to all field names |
 | `field_name_transform` | `transform_func`, `mapping` | Case transformation (see below) |
 | `drop_detection_item` | — | Remove matching detection items |
-| `add_condition` | `conditions: {k: v}`, `negated` (default: `false`) | Inject additional detection conditions |
+| `add_condition` | `conditions: {k: v | [v1, v2, ...]}`, `negated` (default: `false`) | Inject additional detection conditions; list values are OR-linked |
```
mostafa: The bare `|` inside the cell splits the row into extra columns, so this line renders broken. It needs the escaped form the neighboring `field_name_mapping` and `map_string` rows already use.

```suggestion
| `add_condition` | `conditions: {k: v \| [v1, v2, ...]}`, `negated` (default: `false`) | Inject additional detection conditions; list values are OR-linked |
```

crates/rsigma-eval/src/pipeline/parsing.rs
```diff
@@ -791,27 +791,30 @@ fn parse_string_or_list_mapping(
...
-                        }
+                let values = match v {
+                    // YAML sequence → Vec<SigmaValue> (OR semantics, like pySigma)
+                    yaml_serde::Value::Sequence(seq) if seq.is_empty() => {
+                        return Err(EvalError::InvalidModifiers(format!(
+                            "add_condition: empty sequence for field '{key}' is not allowed; \
+                             a detection typo that becomes \"match everything\" is the wrong \
+                             failure direction for a detection engine"
+                        )));
+                    }
```
mostafa: The message carries the rationale from my review comment into a user-facing string. Every other error in this file is a terse statement of what is wrong, and an operator hitting this needs the field name, not the argument for why the check exists. The test only asserts on `empty sequence` and `field_a`, so it still passes.

The comment on the first line also duplicates the one below it and no longer describes the arm it sits above.

```suggestion
                    yaml_serde::Value::Sequence(seq) if seq.is_empty() => {
                        return Err(EvalError::InvalidModifiers(format!(
                            "add_condition: empty sequence for field '{key}'"
                        )));
                    }
```

crates/rsigma-eval/src/pipeline/parsing.rs
```diff
@@ -791,27 +791,30 @@ fn parse_string_or_list_mapping(
...
+                    // YAML sequence → Vec<SigmaValue> (OR semantics, like pySigma)
+                    yaml_serde::Value::Sequence(seq) => {
+                        seq.iter().map(SigmaValue::from_yaml).collect()
                     }
-                    yaml_serde::Value::Bool(b) => SigmaValue::Bool(*b),
-                    yaml_serde::Value::Null => SigmaValue::Null,
-                    _ => SigmaValue::Null,
+                    // Scalar → single-element Vec
+                    other => vec![SigmaValue::from_yaml(other)],
```
mostafa: Worth a look now that the local helper is gone: `SigmaValue::from_yaml` falls back to a debug-formatted string for non-scalar nodes, where the old helper produced `Null`. I checked what that yields in practice, and

```yaml
conditions:
  EventID:
    nested: 1
```

now parses successfully into the literal string value `Mapping {"nested": Number(1)}`.

It will never match anything real, so this is the same class of problem as the empty sequence you just fixed: a malformed pipeline loads clean and the condition quietly does nothing. Rejecting non-scalar nodes in this arm would make the two cases behave consistently.

docs/content/reference/builtin-pipelines.md
```diff
@@ -77,14 +77,18 @@ Each transformation is `add_condition` with a matching `rule_conditions: logsour
...
 | `file_delete` | `EventID: 23` |
 | `clipboard_capture` | `EventID: 24` |
 | `process_tampering` | `EventID: 25` |
 | `file_delete_detected` | `EventID: 26` |
 | `file_block_executable` | `EventID: 27` |
+| `file_block_shredding` | `EventID: 28` |
 | `file_executable_detected` | `EventID: 29` |
+| `sysmon_status` | `EventID: [4, 16]` |
+| `registry_event` | `EventID: [12, 13, 14]` |
+| `wmi_event` | `EventID: [19, 20, 21]` |
```
mostafa: Two things. `sysmon_error` (`EventID: 255`) is in the pipeline now but missing from this table.

And these three rows are appended after `file_executable_detected` (29) rather than slotted in by EventID, which reads oddly given `sysmon_builtin_eventid_order` now enforces exactly that ordering on the YAML this table documents. `sysmon_status` belongs after `network_connection`, `registry_event` after `file_event`, and `wmi_event` after `pipe_created`.

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to RSigma are documented in this file. Each entry correspond
...
+  - `sysmon_status` (EventID 4, 16), `registry_event` (12, 13, 14), `wmi_event` (19, 20, 21),
+    `file_block_shredding` (28) — new categories added.
+  - `pipe_created` updated to include EventID 18 (missing).
+  - **Behavior change**: rules with `category: registry_event`, `wmi_event`, or `sysmon_status`
+    previously got no `EventID` condition at all and matched every Sysmon event; they now match
+    only the relevant EventIDs. Rules with `category: pipe_created` now also match EventID 18.
+- `[]` (empty sequence) is now a parse error instead of silently skipping, preventing pipeline
+  typos from widening rule scope.
+- New tests: `parse_add_condition_with_list_values`, `parse_add_condition_with_empty_sequence`,
+  `test_pipeline_add_condition_list_values_e2e`.
```
mostafa: Three small things.

The entry is hard-wrapped at around 100 columns. The repo convention is not to wrap lines in docs, and every other entry in this file keeps a bullet on one line.

`sysmon_error` (255) is missing from the category list even though the pipeline now adds it.

The last bullet lists test function names, which is not release-note material and is already out of date (the negated multi-value and builtin routing tests are not in it). I would drop it. The `[]` bullet can fold into the first one as well, since "instead of silently skipping" describes a state that never shipped.

mostafa (changes requested):
Thanks, this addressed everything from the last pass. Confirmed locally: `cargo fmt --check` is clean and the Format job is green, the empty sequence now errors at load time and names the field, `yaml_to_sigma_value` is gone in favor of `SigmaValue::from_yaml`, the `wmi_event` block is back in EventID order with `sysmon_error` added, and the negated multi-value plus builtin routing tests cover the gaps I pointed at. Docs build, corpus regression, and all three test jobs are green.

Two things are still blocking.

**Clippy is red.** The new `sysmon_eventid_route` helper trips `collapsible_if` five times (CI is on clippy 1.97). A let-chain fixes it, and the crate already uses them elsewhere.

**The README table row breaks.** The `|` inside the new `add_condition` cell is unescaped, so the row renders with extra columns. The `field_name_mapping` and `map_string` rows right next to it already use the escaped form.

Beyond those, one consistency gap worth closing while you are in here: now that `[]` is a hard error, a non-scalar node is the one remaining path where a malformed pipeline loads clean and the condition quietly does nothing. Details inline. The rest are docs and changelog cleanups.

crates/rsigma-eval/src/pipeline/builtin.rs
mostafa: Fixed in 50e02e65. The chain is a let-chain now, and both tests share an `event_ids` helper instead of repeating the `filter_map`.

crates/rsigma-eval/src/pipeline/builtin.rs
mostafa: Also handled in 50e02e65: `prev_min` and `next_min` are inlined into the format string, and the `Vec` arguments stay positional since they are index expressions.

crates/rsigma-eval/README.md
mostafa: Fixed in b61a018b.

crates/rsigma-eval/src/pipeline/parsing.rs
mostafa: Fixed in 4e0b23c6. The message is now `add_condition: empty sequence for field '{key}'` and the reasoning moved into the comment above the arm. The duplicated comment is gone and both assertions in the test still hold.

crates/rsigma-eval/src/pipeline/parsing.rs
mostafa: Fixed in ba6337b3. A `scalar_condition_value` helper now gates every value, so a mapping or a sequence nested inside a sequence fails at load time with `add_condition: non-scalar value for field '<name>'` instead of loading as a debug string. `parse_add_condition_with_non_scalar_value` covers both shapes.

docs/content/reference/builtin-pipelines.md
mostafa: Fixed in f5b65463. `sysmon_error` is listed and the three rows are slotted in by EventID, so the table now reads in the same order as the pipeline file.

CHANGELOG.md
mostafa: Fixed in e58f6f30. Bullets are unwrapped to match the rest of the file, `sysmon_error` is in the category list, the non-scalar rejection is folded into the empty-sequence bullet, and the test-name bullet is gone.

mostafa (approved):
@frack113 Thank you for your contribution. LGTM. 🚀

mostafa (approved):
@frack113 Thank you for your contribution. LGTM. 🚀

PR #403, opened 2026-07-27
feat(rstix): OASIS STIX 2.1 Interop golden test harness

SecurityEnthusiast:
## Summary

Adds the **OASIS STIX 2.1 Interoperability** (`stix-2.1-interop-v1.0-csd01`) **golden test harness** for rstix: manifest-driven traceability, provenance-aware fixture loading, validation overlay on `Validator::interop_strict()`, bundle-closure checking with TLP exemption, subset containment assertions, certification report generation (`target/interop-report/`), and a silent-skip sentinel.

This PR delivers the **harness infrastructure only** — not OASIS self-certification across all 21 use cases. Normative test-case fixtures and full §2.3 verification land in follow-up work.

## Delivered components

| Component | Path | Role |
|-----------|------|------|
| Interop test target | `tests/interop/main.rs` | Custom runner (`harness = false`; `validate` + `marking` + `graph`) |
| Harness | `tests/interop/harness/` | Manifest, fixture loader + provenance, profile overlay (`STIX-I0002` SHOULD downgrade), closure, containment, certification |
| Cross-cutting checks | `tests/interop/common/` | §2.3 harness smoke until normative fixtures land |
| Manifest | `tests/fixtures/interop/manifest.toml` | `req_id` → `test_id` → fixture → checklist row |
| Smoke fixture | `testcases/harness/tc-harness-closure-smoke.json` | Reference-closed bundle with provenance sidecar |
| Silent-skip guard | `tests/interop_sentinel.rs` | Detects when interop features are disabled at compile time |
| Self-registration | `linkme` | Manifest ↔ test inventory drift detection (both directions) |
| Documentation | `crates/rstix/README.md` | Run commands, report semantics, sentinel behavior, CI job template |

## What the harness verifies today

| Category | Count | Notes |
|----------|-------|-------|
| Harness infrastructure (TESTED → Pass) | 8 | Manifest, fixtures, closure, containment, profile, `interop_strict` smoke bundle |
| §2.3 harness smoke (HARNESS_SMOKE) | 4 | Directory layout, TLP whitelist, identity shape, relationship JSON shape — **not** OASIS verification |
| §4.2 checklist rows (REPORT_ONLY) | 8 | Accounted for in manifest; pending Enrichment + Testing |
| Known OASIS defects (BLOCKED) | 2 | §3.12 Producer defect 19, §3.16 Producer defect 16 |

`summary.json` reports an explicit disposition breakdown. **8 requirements verified** (TESTED + Pass), **4 harness smoke**, **7 report-only placeholders**, **2 blocked** — not “21/21 covered.”

## Test plan

- [*] Workspace fmt / clippy / test / MSRV / lean-features
- [*] Interop harness passes under default parallel workspace test (no `--test-threads=1`)
- [*] Certification report: `requirements_verified == manifest_rows_by_disposition.tested` and smoke rows executed
- [*] Sentinel: passes under `--all-features`
- [*] `cargo deny`: no unmaintained direct dev-deps (`paste` replaced with `pastey`


crates/rstix/tests/interop/harness/certification.rs
```diff
@@ -0,0 +1,327 @@
...
+    let covered = manifest
+        .in_scope_rows()
+        .filter(|row| row_is_covered(row, &recorded))
+        .count();
+
+    assert_eq!(
+        covered, in_scope,
+        "coverage gate: {covered}/{in_scope} in-scope manifest rows covered"
+    );
+}
```
mostafa, line 121: This makes the interop target intermittently fail CI.

`zzz_interop_certification_finalize` only runs after the requirement tests when `--test-threads=1` is passed, but the repo CI (and this PR's own test plan) runs `cargo test --workspace --all-features --locked` without it. Reproduced locally:

- first `cargo test -p rstix --all-features` run: FAILED with 3 failures
- forced concurrency (`-- --test-threads=16`): fails 2 out of 5 runs
- the interop target alone at default threads: 8/8 green, and the documented explicit command is 17/17 green

So it is a load-dependent flake rather than a clean deterministic failure, which is the worst kind to leave in CI.

The failure also amplifies itself: `verify_coverage` asserts while still holding the `OUTCOMES` mutex guard, so the panic poisons the mutex and any still-running tests then die in `record_outcome` with `PoisonError`. One root cause became three reported failures:

```
zzz_interop_certification_finalize panicked: testable manifest requirements without recorded outcomes: ["REQ-HARNESS-MANIFEST-01", "REQ-HARNESS-MANIFEST-02"]
checklist_rows_are_unique panicked: interop outcomes lock: PoisonError
manifest_parses_and_validates panicked: interop outcomes lock: PoisonError
```

Cleanest fix is to stop expressing "run everything, then summarize" as a `#[test]` that must sort last. The harness already has `linkme` self-registration (`INTEROP_TEST_DESCRIPTORS`), so a `harness = false` custom main could iterate the descriptors and finalize deterministically, which removes the `--test-threads=1` requirement entirely. Failing that, mark the finalize `#[ignore]` so only the explicit interop invocation runs it. Either way, drop the guard before asserting so a failure cannot poison the mutex.

Separately on coverage semantics: `row_is_covered` returns `true` unconditionally for `ReportOnly` and `ApiSurface`, and `Blocked` counts as covered whenever it merely names a `blocked_by`. So "21 in scope, 21 covered (100%)" means "every manifest row is accounted for", not "21 requirements verified". Worth making that unambiguous in `summary.json` and the README, since this artifact is what someone would point at as certification evidence.

crates/rstix/tests/interop/common/relationships.rs
```diff
@@ -0,0 +1,29 @@
...
+pub fn assert_relationship_module_ready() {
+    let sample = serde_json::json!({
+        "type": "relationship",
+        "id": "relationship--bc7677e3-3418-4d3f-87e2-66b4b5b7b9ba",
+        "relationship_type": "related-to",
+        "source_ref": "malware--11111111-1111-4111-8111-111111111111",
+        "target_ref": "indicator--22222222-2222-4222-8222-222222222222"
+    });
+    assert_relationship_shape(&sample);
+}
```
mostafa, line 29: This records REQ-2.3-X-08 as Pass on evidence that never exercises rstix: the test builds a hardcoded JSON literal and then asserts that literal contains the fields it just wrote. It would pass identically if rstix were removed from the crate.

Same pattern elsewhere: REQ-2.3-X-09 is a directory-existence check (`common/gating.rs`) and REQ-2.3-X-04 asserts a constant list has 9 entries (`common/bundle_closure.rs`).

These are perfectly reasonable as harness smoke tests. The issue is that they feed a certification artifact intended for OASIS SXP/SXC self-certification while the 44 normative OASIS fixtures are explicitly not in this PR. Suggest dispositioning those rows as blocked/pending-fixtures rather than passing requirements, so `summary.json` does not overstate conformance until the real fixtures land.

mostafa (commented):
Reviewed the interop harness. The backbone is good: manifest-driven traceability with bidirectional drift detection (missing tests and orphan tests), the provenance-sidecar requirement on fixtures, and the silent-skip sentinel are all the right ideas. `clippy --all-features --all-targets -- -D warnings` and `fmt` are clean, and the documented explicit run is 17/17 green.

One CI blocker though: the certification finalize test races and intermittently fails under concurrency, because it depends on running last and CI does not pass `--test-threads=1`. Reproduced below. There is also a smaller honesty concern about requirement rows recorded as Pass on evidence that does not exercise rstix. Details inline.

SecurityEnthusiast:
# Fixes:
1. **CI flake:** replaced parallel `#[test]` + `zzz_finalize` with custom `main()` runner — tests run sequentially, finalize always last.
2. **cargo deny:** replaced unmaintained `paste` with `pastey`.
3. **Honesty:** §2.3-X-04/05/08/09 disposition → `HARNESS_SMOKE`; certification records `HARNESS_SMOKE` outcome, not `Pass`.
4. **Docs:** README + CI gate clarify `covered_requirements` (accounted for) vs `requirements_verified` (TESTED + Pass).

mostafa (approved):

PR #388, opened 2026-07-24
feat(rstix): close STIX 2.1 wire conformance gaps at parse and interop_strict

SecurityEnthusiast:
## Summary

This PR closes the remaining documented STIX 2.1 wire conformance gaps in `rstix`: normative **MUST** rules are enforced at **T0 parse** (`serde`), **SHOULD** checks stay on **`Bundle::validate()`** (T1), and **`Validator::interop_strict()`** (T2) applies structured diagnostics with **spec-correct severities**. It also fixes places where the first draft was wrong vs OASIS — rejecting open-vocabulary strings at parse (PR review), treating custom open-vocab values as `interop_strict` failures, and leaving three Data Model items incorrectly listed as deferrals. Stale “conformance exception” documentation is replaced with accurate wire-conformance and feature-oriented docs.

### T0 parse (`ModelError` / `ParseError`)

- **Non-empty strings** — SDO `name`, grouping `context`; non-empty `object_refs` on report, grouping, note, and opinion.
- **Reference kinds** — `object_refs` and `language-content.object_ref` accept any STIX Object (SDO, SCO, SRO, Meta), not only SDO/SCO.
- **Wire formats** — IPv4/IPv6/MAC address syntax; full RFC 3986 URL parse (removed `http`/`https`/`ftp` scheme whitelist); hash map keys in `HASH_ALGORITHM_ENUM` or `x_` extension form with known-algorithm value formats.
- **Closed enumerations at parse** — artifact `encryption_algorithm` (was T1 advisory); `OpinionValue`; extension types; aligned with STIX §10.4 / closed vocab tables.
- **Open vocabulary at parse (§2.14 — corrected after review)** — `grouping.context`, `malware-analysis.result`, and `windows-pebinary-ext.pe_type` **accept any string** at parse; removed `validate_open_vocab_string()` from those types and from `model/validate.rs`. Non-empty grouping `context` remains a T0 MUST.
- **Time ordering** — `malware-analysis` `start_time` / `end_time` / `result_name` ordering enforced in `validate()`.
- **§7.1.1 language-content MUST** — `object_modified` mismatch and translation mirroring (type, list length, object shape) for **existing** target fields rejected at parse; keys for properties absent on the target **silently ignored** (no error, no warning).
- **§7.2.3 granular selectors MUST** — selectors must resolve on object wire JSON at bundle parse (`validate_granular_markings_semantics()`), including custom objects when `allow_custom` is enabled.
[...]

PR #410, opened 2026-07-29
rstix: OASIS STIX 2.1 Interop normative fixtures and §2.3 suite checks

SecurityEnthusiast:
## Summary

Extends the OASIS STIX 2.1 Interop harness landed in #403 with normative test-case fixtures, suite-wide §2.3 cross-cutting checks, and the parse/validate hooks they require. 

### Fixtures and inventory

- Adds normative OASIS interop JSON under `tests/fixtures/interop/testcases/` with `.provenance.toml` sidecars.
- **44/44** gating fixture inventory: **42** walkable in the suite-wide fixture walk; **2** `BLOCKED` on unrepairable OASIS §9.1 publisher defects (§3.12 malware id defect 19, §3.16 report `published` defect 16).
- Covers §3.1–§3.21 use-case fixtures incrementally (attack-pattern, campaign, confidence, course-of-action, TLP, grouping, infrastructure, intrusion-set, indicator, location, malware, malware-analysis, note, observed-data, opinion, report, sighting, tool, versioning, vulnerability).
- Synthetic helpers under `testcases/harness/` and `testcases/common/` stay excluded from the normative walk.

### §2.3 suite-wide execution

- Promotes **18 §2.3 manifest rows** from harness-smoke layout to **`disposition = TESTED`**, executed over all walkable normative fixtures.
- New `tests/interop/common/`: `fixture_catalog`, `fixture_walk`, `wire_preservation`, expanded `producer`, `consumer`, `identity`, `relationships`, `sco`, `validation`, `bundle_closure`.
- New `tests/interop/harness/interop_gate.rs` and `use_case.rs`: interop bundle gate (reference closure, `interop_bundle_strict`, overlay), per-type producer use-case rules stricter than STIX §4.x where §3.x requires it, wire subset preservation on re-serialize.
- Certification report still written to `target/interop-report/` (`summary.json`, traceability CSV). **`TESTED` + `Pass` means the harness recorded success — not OASIS SXP/SXC certification.**

### Public API / model changes

- **`ParseOptions::interop_bundle()`** and **`exempt_predefined_tlp_marking_refs`**: interop TLP marking ref exemption at parse/reference check.
- **`Validator::interop_bundle_strict()`**: bundle-scoped strict profile for the harness.
- **`Grouping.name`**: optional at parse (STIX §4.4); interop producer rules still require `name` where the use case mandates it.

### Docs

- `CHANGELOG.md`: new `[Unreleased]` entry for this work; **#403 entry left unchanged**.
- `crates/rstix/README.md`: interop inventory, suite-wide §2.3 behavior, report field semantics.

### Explicitly out of scope

- OASIS certification / full ~354-row manifest coverage
- Dedicated CI interop report gate job (documented in README as template only)
- `tests/interop/use_cases/` per-use-case Producer/Consumer modules (future increment)

## Test plan

Workspace gate:

- [x] `cargo check -p rstix --no-default-features --locked`
- [x] `cargo clippy -p rstix --no-default-features --all-targets --locked -- -D warnings`
- [x] `cargo deny check`
- [x] `cargo audit`

rstix interop and validation:

- [x] `cargo test -p rstix --test interop_sentinel --features validate,marking,graph --locked`
- [x] `cargo test -p rstix --test validate_conformance --features validate --locked`
- [x] `cargo test -p rstix --test validate_diagnostic_coverage --features validate --locked`
- [x] `cargo test -p rstix --test validate_pipeline --features validate --locked`
- [x] `cargo test -p rstix --test spec --features serde --locked`
- [x] `cargo test -p rstix --test bundle --features serde --locked`
- [x] `cargo test -p rstix --test validation --features validate --locked`
- [x] `cargo test -p rstix --features graph --test graph --locked`
- [x] `cargo test -p rstix --features marking --test marking --locked`
- [x] `cargo test -p rstix --features taxii --test taxii_client --locked`
- [x] `cargo test -p rstix --features taxii-store --test taxii_store --locked`

Docs (CHANGELOG changed):

- [x] `cd docs && npm ci && npm run docs:build && npm run docs:validate`

Optional (requires local ATT&CK bundle):

- [*] `RSTIX_ATTCK_BUNDLE=… cargo test -p rstix --features serde attck_corpus_roundtrip_when_present -- --nocapture`


crates/rstix/tests/interop/common/fixture_walk.rs
```diff
@@ -0,0 +1,74 @@
...
+        return;
+    }
+    for entry in std::fs::read_dir(dir).expect("read dir") {
+        let entry = entry.expect("dir entry");
+        let path = entry.path();
+        if path.is_dir() {
+            walk_dir(&path, f, filter);
+        } else if path.extension().is_some_and(|ext| ext == "json") {
+            let sidecar = path.with_extension("provenance.toml");
+            if !sidecar.exists() {
```
mostafa, line 118: The suite-wide fixture inventory can silently shrink to nothing while every §2.3 row still reports `TESTED` + `Pass`.

Any `.json` whose `.provenance.toml` sidecar is missing is skipped here, and nothing anywhere in the harness asserts how many fixtures the walk actually visited. I grepped the whole interop tree for a count assertion and found none. So deleting a fixture, or merely renaming or losing a sidecar, quietly drops it from all 18 suite-wide rows. The suite still passes, and the certification report continues to describe a 44-fixture inventory that no longer exists.

That matters more here than in an ordinary test suite, because the entire value of this harness is traceable coverage. A silent-shrink path means the report can overstate what was exercised without anything failing.

This is the same drift class #403 already guards against for manifest rows versus registered tests, just one level further down, so the fix can take the same shape: assert the walked count against a checked-in expected value (44 total, 42 walkable) and fail in both directions, so an added fixture that nobody registered fails too. Deriving the expectation from the manifest inventory rather than a bare constant would be better still.

A missing sidecar is also worth distinguishing from a deliberate exclusion. Right now it's indistinguishable from `harness/` and `common/` synthetics, which are excluded on purpose; a missing sidecar on a normative fixture is closer to a defect and could reasonably panic rather than `continue`.

mostafa (commented):
This is a strong increment over #403, and the two things I expected to be problems are both handled honestly.

The fixture repairs are real PDF-extraction artifacts (escaped underscores, smart quotes, missing commas), each classified `MECHANICAL` or `JUDGMENT`, with published-doc defects recorded as `divergence_recorded` rather than edited away. The overlay downgrades are all SHOULD-level, scoped to the harness overlay rather than the default validator, and disclosed in the report's "SHOULD-level downgrades" section. The two `BLOCKED` fixtures carry their defect numbers instead of being quietly passed. And the promoted rows are no longer vacuous: `wire_preservation` re-serializes every parsed object and asserts no meaningful wire field is dropped across all 42 walkable fixtures, which is genuine work.

Making `Grouping.name` optional is a real bug fix, not a test-driven loosening. STIX §4.4 lists only `context` and `object_refs` as required for grouping, unlike `report` where `name` is required, so the parse-time requirement added in #388 was over-strict. A normative fixture failing to parse is exactly what this harness exists to catch, and it caught it.

Two things to address, one before merge:

1. That change is breaking and isn't labeled as such. rstix has shipped through 0.20.0, so `pub name: String` to `pub name: Option<String>` breaks every construction and read site at compile time.
2. The fixture inventory can silently shrink while the suite stays green and the report keeps claiming 44 fixtures.

Gates all pass locally: `fmt`, `clippy --all-features --all-targets -D warnings`, `clippy --no-default-features`, the interop target, and the full `--all-features` test run.

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to RSigma are documented in this file. Each entry correspond
 
 ## [Unreleased]
 
+### rstix: OASIS STIX 2.1 Interop normative fixtures and §2.3 suite checks (#410)
+
+- Adds normative OASIS interop test-case JSON under `tests/fixtures/interop/testcases/` with provenance sidecars for the §2.3 cross-cutting and §3.x use-case inventory (44 gating fixtures; 42 exercised in the suite-wide walk; 2 recorded as `BLOCKED` on unrepairable OASIS §9.1 publisher defects).
+- Runs §2.3 cross-cutting manifest rows suite-wide over walkable fixtures through the interop bundle gate, per-type producer use-case rules, wire preservation on re-serialize, and expanded consumer, identity, relationship, and SCO checks.
+- **`ParseOptions::interop_bundle()`** and **`exempt_predefined_tlp_marking_refs`** extend bundle parse and reference validation for interop TLP marking fixtures; **`Validator::interop_bundle_strict()`** adds a bundle-scoped strict profile for the harness.
+- **`Grouping.name`** is optional at parse, matching STIX §4.4.
```
mostafa: Please mark this as a breaking change. rstix has shipped through 0.20.0, and `pub name: String` to `pub name: Option<String>` is a source break for any caller that constructs a `Grouping` or reads `.name` without unwrapping.

The spec rationale in this bullet is correct and worth keeping. The gap is that a reader scanning release notes for upgrade impact sees a conformance fix and nothing signalling that their code stops compiling. Since `CHANGELOG.md` is the source of truth for the release body, that omission propagates straight into the published notes.

A short breaking marker plus the one-line migration (`name` now yields `Option<String>`) is enough.

crates/rstix/src/validate/checks/references.rs
```diff
@@ -16,12 +17,22 @@ use super::super::wire::{
 };
 use super::{ValidationContext, ValidationReport};
 
+fn is_marking_ref_property(path: &str) -> bool {
+    path.contains("object_marking_refs") || path.contains("marking_ref")
```
mostafa: Minor: substring matching makes this looser than intended. A custom property such as `x_object_marking_refs_note`, or any path segment that merely contains `marking_ref`, satisfies this and becomes eligible for the exemption.

Severity is low, since a match additionally requires the opt-in `exempt_predefined_tlp_marking_refs` flag and a target that is a predefined TLP id, so the blast radius is confined to the harness profile today. Still, matching the final path segment exactly is both cheaper to reason about and immune to custom properties that happen to embed these names.

crates/rstix/tests/interop/common/wire_preservation.rs
```diff
@@ -0,0 +1,78 @@
+//! Wire → parse field preservation for §2.3 P-03 / C-05.
+
+use rstix::core::StixId;
+use rstix::model::Bundle;
+use serde_json::Value;
+
+use crate::harness::containment::assert_json_contains;
+
+const SKIP_WIRE_COMPARE: &[&str] = &["spec_version"];
```
mostafa, line 11: Worth a one-line comment on why `spec_version` is exempt from wire comparison. I assume rstix normalizes it on parse, so a literal comparison would be a false positive, but as written this reads like a convenience skip that was added to make a fixture pass.

Given the point of this module is to prove no field is silently dropped, an unexplained exemption is the one thing a reader will be suspicious of, and this list is exactly where future exemptions will accumulate. Stating the rule for what may legitimately be added keeps that pressure in check.

SecurityEnthusiast:
> This is a strong increment over #403, and the two things I expected to be problems are both handled honestly.
> 
> The fixture repairs are real PDF-extraction artifacts (escaped underscores, smart quotes, missing commas), each classified `MECHANICAL` or `JUDGMENT`, with published-doc defects recorded as `divergence_recorded` rather than edited away. The overlay downgrades are all SHOULD-level, scoped to the harness overlay rather than the default validator, and disclosed in the report's "SHOULD-level downgrades" section. The two `BLOCKED` fixtures carry their defect numbers instead of being quietly passed. And the promoted rows are no longer vacuous: `wire_preservation` re-serializes every parsed object and asserts no meaningful wire field is dropped across all 42 walkable fixtures, which is genuine work.
> 
> Making `Grouping.name` optional is a real bug fix, not a test-driven loosening. STIX §4.4 lists only `context` and `object_refs` as required for grouping, unlike `report` where `name` is required, so the parse-time requirement added in #388 was over-strict. A normative fixture failing to parse is exactly what this harness exists to catch, and it caught it.
> 
> Two things to address, one before merge:
> 
> 1. That change is breaking and isn't labeled as such. rstix has shipped through 0.20.0, so `pub name: String` to `pub name: Option<String>` breaks every construction and read site at compile time.
> 2. The fixture inventory can silently shrink while the suite stays green and the report keeps claiming 44 fixtures.
> 
> Gates all pass locally: `fmt`, `clippy --all-features --all-targets -D warnings`, `clippy --no-default-features`, the interop target, and the full `--all-features` test run.

Thank you for the detailed review. Both pre-merge items are addressed in the latest push.

1. **`Grouping.name` breaking change** — The `[Unreleased]` entry now has an explicit **Breaking:** bullet: `Grouping.name` is `Option<String>` at parse, with a one-line migration note for callers that constructed or read it as `String`. The STIX §4.4 rationale is unchanged.

2. **Fixture inventory shrink** — `certification::finalize()` now calls `assert_fixture_inventory_counts()`, which pins **44** gating fixtures, **42** suite-walk fixtures, and **2** blocked fixtures, and fails if any count drifts in either direction. A missing `.provenance.toml` on a normative testcase JSON now **panics** instead of being skipped silently (`testcases/harness/missing-sidecar.json` remains the only deliberate exception). Fixture paths are normalized to forward slashes so suite-wide walks behave the same on Windows.

Also addressed the minor items: exact leaf-property matching for TLP marking-ref exemption, and a comment on the `spec_version` wire-compare skip. Interop target and clippy pass locally.

crates/rstix/tests/interop/common/fixture_walk.rs
SecurityEnthusiast: Fixed in the latest push.

- `assert_fixture_inventory_counts()` runs during `certification::finalize()` and asserts **44** gating, **42** suite-walk, and **2** blocked fixtures — failure in **both** directions (shrink or growth).
- Missing `.provenance.toml` on normative testcase JSON now **panics**; only `testcases/harness/missing-sidecar.json` is exempt.
- `normalize_fixture_path()` ensures prefix checks work on Windows (this also fixed the `data-markings` suite walk failure on `windows-latest`).

CHANGELOG.md
SecurityEnthusiast: Done. The bullet is now prefixed with **Breaking:** and notes that `Grouping.name` is `Option<String>` at parse, with migration guidance for construction and reads. The STIX §4.4 conformance rationale stays in the same bullet.

crates/rstix/src/validate/checks/references.rs
SecurityEnthusiast: Fixed. `is_marking_ref_property` now matches the path **leaf segment** exactly (`object_marking_refs` or `marking_ref`, including indexed paths like `object_marking_refs[0]`) instead of substring-matching anywhere in the path. The exemption still requires `exempt_predefined_tlp_marking_refs` plus a predefined TLP id.

crates/rstix/tests/interop/common/wire_preservation.rs
SecurityEnthusiast: Added a comment above `SKIP_WIRE_COMPARE`: `spec_version` is injected/normalized on parse when absent on the wire, so a literal comparison would false-fail without indicating a real field drop. Only properties with a documented normalization rule belong on that list.

mostafa (approved):

PR #412, opened 2026-07-29
Use `jemalloc` for better scaling performance with `musl`

alltilla:
## Summary

- use jemalloc as the global allocator on musl targets, restoring the multicore scaling the released image had lost
- add `scripts/perf/image-compare.sh`, which measures two container images end-to-end and can see artifact-level cost the native harnesses cannot

The released `ghcr.io/timescale/rsigma` binary is static musl, and musl's mallocng serializes the concurrent allocation `Engine::evaluate_batch` fans across rayon. The result is that the published artifact did not scale: with `--logsource-routing` over the pinned corpus the daemon reached 53,435 events/s while the same binary evaluated 57,463 events/s on a single core. The fan-out returned less than nothing.

This is invisible to every existing benchmark, because they all measure a native build against the system allocator, which does not have the problem. `BENCHMARKS.md` reports the daemon at 5.8x its single-core rate on the raw-blob lane; the shipped image was at 0.93x.

## Measurements

Host: AMD Ryzen 7 PRO 6850U, 8 physical cores / 16 logical CPUs, Linux. Corpus: pinned SigmaHQ `994da166` (3,132 rules), `raw_windows` lane, `--batch-size 512`, `/dev/null` sink.

- **Single-core** is `scripts/perf/baseline-eval.sh` with `REPEAT=5`, net of rule load. `engine eval` is single-threaded, so this is a per-core rate.
- **Daemon** is the median of three 30 s runs of the new `image-compare.sh`. The daemon container is pinned to 12 logical CPUs (6 physical cores) and the k6 container to the remaining 4 logical CPUs (2 physical cores), so the load generator cannot starve the engine.
- **Daemon CPU** is logical-CPU equivalents sampled from `docker stats` (100% = one logical CPU), against the 12 the daemon was allowed.
- **Parallel speedup** is Daemon / Single-core: what the fan-out actually bought.

Both columns run against binaries extracted from a base-revision image and a head image, so nothing but the allocator differs.

| Variant | Build | Single-core (ev/s) | Daemon (ev/s) | Daemon CPU | Parallel speedup |
|---|---|---:|---:|---:|---:|
| default | musl | 26,557 | 36,875 | 8.0 | 1.39x |
| default | musl + jemalloc | 30,045 | 95,505 | 6.7 | 3.18x |
| `--logsource-routing` | musl | 57,463 | 53,435 | 8.0 | **0.93x** |
| `--logsource-routing` | musl + jemalloc | 65,826 | 160,238 | 6.8 | 2.44x |

Head against base, same row pairs:

| Variant | Daemon before | Daemon after | Change |
|---|---:|---:|---:|
| default | 36,875 | 95,505 | **2.59x** |
| `--logsource-routing` | 53,435 | 160,238 | **3.00x** |

Three things the numbers say:

1. Stock musl at **0.93x** parallel speedup means six physical cores delivered less than one. The allocator was not merely a tax on the fan-out, it consumed all of it.
2. Single-core throughput improves only ~14% (26,557 to 30,045; 57,463 to 65,826). The gap between ~14% single-threaded and ~3x on the daemon is what identifies the cost as allocator contention rather than per-event work.
3. jemalloc reaches the higher figures on *less* CPU, 6.7-6.8 against 8.0.

Match counts are identical in every cell (151,185 default, 1,120 routing), so this is a throughput change and not a behaviour change. Allocation-heavy rule loading drops from 0.86 s to 0.49 s. The stripped binary grows from 42.5 MB to 43.3 MB.

For reference, a glibc build of the same revision lands within ~4% of musl + jemalloc, which is what suggested the allocator in the first place.

## Scope

The dependency sits under `[target.'cfg(target_env = "musl")'.dependencies]` rather than behind a feature, so glibc and macOS keep the system allocator and neither `cargo install` nor a native `cargo build` changes behaviour. `jemalloc-sys` compiles C, so the Docker builder gains `build-base` alongside `musl-dev`.

## On the new perf gates

The regression gates added in #409 build natively on the CI runner, so they are glibc and cannot observe this class of change in either direction. That is how a 3x gap survived an otherwise thorough benchmarking effort. Adding a musl lane to the gate would close the blind spot; happy to follow up separately, since it is a different concern from this change.

`image-compare.sh` needs only Docker, git and python3 (no Rust toolchain, no local k6) and reuses `daemon-load.js` unchanged. It reports a median of three runs and accepts cpusets for both sides, which matters more than it sounds: unpinned runs on a loaded host swung by 2-3x between identical invocations. It also charges throughput to measured wall time across the post-load drain rather than to the nominal load duration, since events still queued when k6 exits would otherwise inflate the rate.

To reproduce base against head:

```bash
git archive <base-sha> | docker build -t rsigma:base -
BASELINE=rsigma:base DAEMON_CPUS=0-11 LOAD_CPUS=12-15 VUS=16 \
  scripts/perf/image-compare.sh
```

## Limits

- Measured on one x86-64 Linux host. aarch64 musl and macOS are untested, though the mechanism is not architecture-specific.
- jemalloc was measured; mimalloc was not. This is not a claim that jemalloc is the best of the alternatives, only that mallocng is the problem.
- No tests are added. Allocator selection has no branch to assert on, and the behavioural check that matters, identical match counts either side, is covered by the existing differential suites and reproduced above.

## Test plan

All of the below run against the head commit of this branch.

- [x] `cargo test --workspace --all-features --locked` (3,728 passed, 0 failed)
- [x] `cargo doc --workspace --all-features --locked --no-deps` (no warnings)
- [x] `cargo deny check` (advisories, bans, licenses, sources ok)
- [x] `docker build .` with the repo Dockerfile and `--locked` (no warnings; 43.3 MB against 42.5 MB on base)
- [x] `cd docs && npm run docs:build && npm run docs:validate` (115 pages, all internal links valid; CHANGELOG and BENCHMARKS changed, and BENCHMARKS is included into `docs/content/benchmarks.md`)
- [x] `scripts/perf/baseline-eval.sh` and `scripts/perf/image-compare.sh` against base and head images
- [x] `shellcheck scripts/perf/image-compare.sh`

The `postgres_integration` and `cli_daemon_nats` suites need their testcontainers images (`postgres:11-alpine`, `nats:2.10.14`) resolvable; they fail at container start on a host that cannot pull them, independently of this change.


scripts/perf/image-compare.sh
```diff
@@ -0,0 +1,184 @@
...
+set -euo pipefail
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+repo_root="$(cd "${script_dir}/../.." && pwd)"
+fixtures="${1:-${repo_root}/target/perf-fixtures}"
+lane="${2:-raw_windows}"
+shift 2 2>/dev/null || shift $# || true
+
+baseline="${BASELINE:-ghcr.io/timescale/rsigma:0.20.0}"
+k6_image="${K6_IMAGE:-grafana/k6:latest}"
```
mostafa: Please pin k6 to a versioned digest. `grafana/k6:latest` makes comparisons non-reproducible and executes mutable third-party code.

scripts/perf/image-compare.sh
```diff
@@ -0,0 +1,184 @@
...
+        "${image}" engine daemon -r /rules --input http \
+        --api-addr 0.0.0.0:9090 --allow-plaintext \
+        --batch-size "${batch_size}" --output "file:///dev/null" "$@" >/dev/null
+
+    local ready=""
+    for _ in $(seq 1 240); do
+        curl -sf "http://${addr}/readyz" 2>/dev/null | grep -q rules_loaded && { ready=yes; break; }
+        sleep 0.5
+    done
+    [ -n "${ready}" ] || {
```
mostafa, line 141: These match totals are not comparable: each image receives independently randomized batches, and the faster image processes more requests. Please validate matches with a deterministic fixed-input pass or remove this field.

scripts/perf/image-compare.sh
alltilla: Pinned to `grafana/k6:2.1.0@sha256:65c920dc067d5e2e00befbf982af6ad6ad0117034e8b1c65817c7975c52d4669` (current latest stable); `K6_IMAGE` still overrides.

scripts/perf/image-compare.sh
alltilla: Agreed — removed the matches column and its bookkeeping; the report now shows only events/s, cores and eps/core.

mostafa:
@alltilla The rate of change is so fast that the changelog doc gets stale pretty quickly. Please fix it. 🙏

alltilla:
I'm not on computer right now, feel free to update it, you should be able to push on my branch. I can update it in a couple of hours otherwise 🙂 🙏

mostafa:
@alltilla No worries! I'll fix it.

mostafa (approved):
@alltilla Thank you for your contribution. LGTM. 🚀

PR #413, opened 2026-07-29
test(rstix): OASIS §3.1 Attack Pattern Producer and Consumer interop tests

SecurityEnthusiast:
# Summary
- Adds `tests/interop/use_cases/attack_pattern/` for OASIS Interop **§3.1.1–§3.1.7** (Producer `P-01`..`P-13`, Consumer `C-01`..`C-05`, checklist `REQ-CHK-SXP/SXC-3.1`, examples `EX-4.1`/`EX-4.2`/`EX-7.1`), with §3.1.1 as real **TESTED** scope (not REPORT_ONLY).
- Fixtures under `examples/attack-pattern/`; §3.1.4.2 keeps truncated published UUIDs and asserts parse reject (no invented digits).
- Traceability only — **not** OASIS SXP/SXC certification; **1/21** use cases. Interop report expect **51** TESTED / **7** REPORT_ONLY / **2** BLOCKED.

## Test plan
- [*] `cargo test --workspace --all-features --locked`
- [*] `cargo test -p rstix --test interop --features validate,marking,graph --locked` → `tested_rows_passed == 51`
- [*] `cargo +1.88.0 check --workspace --all-targets --all-features --locked`

### DIFF
--- a/crates/rstix/tests/interop/use_cases/attack_pattern/consumer.rs
+++ b/crates/rstix/tests/interop/use_cases/attack_pattern/consumer.rs
@@ -1,0 +1,155 @@
+//! §3.1.5 Required Consumer Persona Support (REQ-3.1-C-01..C-05).
+
+use crate::interop_test;
+use rstix::core::StixId;
+use rstix::model::sdo::{AttackPattern, Identity, Vulnerability};
+use rstix::model::sro::Relationship;
+use serde_json::Value;
+
+use crate::common::fixture_catalog::{
+    parse_fixture_objects, summarize_fixture_wire, use_case_object_ids,
+};
+use crate::common::wire_preservation::{
+    assert_identity_fields_preserved, assert_wire_object_preserved,
+};
+use crate::harness::fixture::load_fixture;
+use crate::harness::interop_gate::validate_interop_fixture;
+use crate::use_cases::attack_pattern::{FIXTURE_TARGETS, PRODUCER_FIXTURES};
+
+fn for_each_attack_pattern_fixture(mut f: impl FnMut(&str)) {
+    for relative in PRODUCER_FIXTURES {
+        f(relative);
+    }
+}
+
+/// REQ-3.1-C-01 — Consumer supports §3.1.2 Producer Persona properties on normative fixtures.
+pub fn assert_supports_producer_props() {
+    for_each_attack_pattern_fixture(|relative| {
+        let fixture = load_fixture(relative);
+        let objects = parse_fixture_objects(&fixture.json)
+            .unwrap_or_else(|err| panic!("{relative}: parse fixture: {err}"));
+        let use_case_ids = use_case_object_ids(relative, &objects);
+        let bundle = validate_interop_fixture(relative, &fixture.json)
+            .unwrap_or_else(|err| panic!("{relative}: interop gate: {err}"));
+
+        for object_id in use_case_ids {
+            let wire = objects
+                .iter()
+                .find(|obj| obj.get("id").and_then(Value::as_str) == Some(object_id.as_str()))
+                .unwrap_or_else(|| panic!("{relative}: wire object {object_id}"));
+            let stix_id = StixId::parse(&object_id).expect("object id");
+            let ap = bundle
+                .get_typed::<AttackPattern>(&stix_id)
+                .unwrap_or_else(|| panic!("{relative}: typed attack-pattern {object_id}"));
+            assert!(!ap.common.external_references.is_empty());
+            assert!(!ap.kill_chain_phases.is_empty());
+            assert_wire_object_preserved(relative, wire, &bundle, &object_id);
+        }
+    });
+}
+
+/// REQ-3.1-C-02 — Consumer receives Identity, Attack Pattern(s), and SROs (§3.1.3.2).
+pub fn assert_receives_triad() {
+    let relative = FIXTURE_TARGETS;
+    let fixture = load_fixture(relative);
+    let summary = summarize_fixture_wire(&fixture.json)
+        .unwrap_or_else(|err| panic!("{relative}: summarize fixture: {err}"));
+    assert_eq!(summary.identity_ids.len(), 1);
+    assert_eq!(
+        summary.primary_sdo_count, 2,
+        "attack-pattern + vulnerability"
+    );
+    assert_eq!(summary.relationship_count, 1);
+
+    let bundle = validate_interop_fixture(relative, &fixture.json)
+        .unwrap_or_else(|err| panic!("{relative}: interop gate: {err}"));
+    for identity_id in &summary.identity_ids {
+        let id = StixId::parse(identity_id).expect("identity id");
+        assert!(
+            bundle.get_typed::<Identity>(&id).is_some(),
+            "{relative}: Identity {identity_id} must parse"
+        );
+    }
+    assert_eq!(bundle.objects_of_type::<AttackPattern>().count(), 1);
+    assert_eq!(bundle.objects_of_type::<Vulnerability>().count(), 1);
+    assert_eq!(bundle.objects_of_type::<Relationship>().count(), 1);
+}
+
+/// REQ-3.1-C-03 — Consumer resolves `created_by_ref` Identity fields (§3.1.3.2).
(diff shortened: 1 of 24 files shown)


mostafa (commented):
This is the cleanest of the interop PRs so far. The structure is honest: 24 test registrations for 24 new `TESTED` rows, so there is no one-test-claims-many-rows inflation, and the 51/7/2 manifest counts match the description exactly. §3.1.1 genuinely earned its promotion out of `REPORT_ONLY`, since it parses a normative fixture through the gate, does a typed `get_typed::<AttackPattern>`, and asserts on the parsed value rather than on prose.

The best thing here is §3.1.4.2: keeping the published truncated UUIDs instead of inventing digits, recording defect 21 in the sidecar, and asserting parse rejects them. That is exactly the right instinct, and it tests rstix rather than the fixture.

The consumer side does real work too: typed lookups, wire preservation, `created_by_ref` resolution, and relationship endpoint typing. Loading examples by explicit const path is also the right call, since a missing fixture panics instead of silently dropping out of a walk.

Four findings, all nit-to-moderate, none blocking. The theme of the first is that three rows are backed by evidence identical to another row, which matters only because traceability is this harness's product. Your doc comments are candid about it, so it reads as a labeling choice rather than concealment.

One item that is **not** this PR: `cargo clippy -p rstix --all-targets --all-features -- -D warnings` exits 101 at the toolchain `rust-toolchain.toml` pins, on three `gating.rs` sites that came in with #410 and are now on main. This branch does not touch that file. Flagging it only because it means `cargo clippy` from a fresh checkout is red for reasons unrelated to your change.

Gates on my end: `fmt` clean, interop target passes, full `rstix --all-features` suite passes, `--no-default-features` clippy clean.

crates/rstix/tests/interop/use_cases/attack_pattern/producer.rs
```diff
@@ -0,0 +1,320 @@
...
+    let identities: Vec<_> = objects
+        .iter()
+        .filter(|obj| obj.get("type").and_then(Value::as_str) == Some("identity"))
+        .collect();
+    assert_eq!(identities.len(), 1, "{relative}: expected one Identity");
+    assert_identity_shape(relative, identities[0]);
+}
+
+/// REQ-3.1-P-04 — Attack Pattern conforms to STIX §4.1 via interop gate on §3.1.3.1.
+pub fn assert_spec_conformance() {
```
mostafa, line 109: P-04 and P-01 are backed by the same assertion: this body is a single call to `assert_create_attack_pattern()`. Two manifest rows, one piece of evidence.

The same pattern appears twice more. `REQ-CHK-SXP-3.1` here and `REQ-CHK-SXC-3.1` in `consumer.rs` have bodies that differ only in panic text, both looping `PRODUCER_FIXTURES` through `validate_interop_fixture`. And C-04's wire-preservation over `FIXTURE_TARGETS` is a strict subset of C-01, which already does the same over both fixtures.

Net effect: of the 24 new `TESTED` rows, about three add no evidence beyond a sibling row, so `requirements_verified` overstates distinct verification by that much. Since the certification report is the deliverable, that number should mean what a reader assumes.

P-04 could earn its own row cheaply. The doc comment says "conforms to STIX §4.1", which is a stronger claim than "the bundle passed the gate": asserting the strict validator emits no MUST-level diagnostics scoped to the attack-pattern object would verify the §4.1 claim distinctly from P-01's "producer content parses". For the SXP/SXC checklist pair, the personas legitimately differ even when the data does not, so recording one as covered-by-the-other, or having the consumer row assert consumption rather than re-validation, would keep the row honest.

crates/rstix/tests/interop/use_cases/attack_pattern/producer.rs
```diff
@@ -0,0 +1,320 @@
...
+    assert_create_attack_pattern();
+}
+
+/// REQ-3.1-P-05 — `type` is `attack-pattern`.
+pub fn assert_prop_type() {
+    let (_ap, _) = load_attack_pattern(FIXTURE_CREATE);
+    assert_eq!(AttackPattern::TYPE_NAME, "attack-pattern");
```
mostafa: This assertion cannot fail from fixture data. `AttackPattern::TYPE_NAME` is a compile-time constant, so comparing it to `"attack-pattern"` is a tautology, and the object loaded on the line above is discarded via `let (_ap, _)`.

The row is not entirely without force, but only incidentally: the `get_typed::<AttackPattern>` inside `load_attack_pattern` panics if the discriminant does not match, so the type really is being checked. The problem is that the explicit assertion contributes nothing while appearing to be the check, which is the same shape as the hardcoded-literal assert in `assert_relationship_module_ready` from #403.

Either assert the parsed or wire `type` of the object you just loaded, or drop the assertion and say in the doc comment that the typed lookup is the proof. The latter is honest and shorter.

crates/rstix/tests/interop/use_cases/attack_pattern/examples.rs
```diff
@@ -0,0 +1,62 @@
...
+            .divergence_recorded
+            .iter()
+            .any(|d| d.defect == 21),
+        "expected defect 21 recorded for truncated UUID ids"
+    );
+    let err = Bundle::parse_with_options(&fixture.json, &ParseOptions::new().interop_bundle())
+        .expect_err("§3.1.4.2 truncated UUID ids must fail parse");
+    let msg = err.to_string();
+    assert!(
+        msg.contains("uuid") || msg.contains("UUID") || msg.contains("id"),
```
mostafa: This predicate is close to unfalsifiable: `msg.contains("id")` matches the word "invalid", which appears in most parse errors, so essentially any failure satisfies it.

That undercuts an otherwise excellent test. The point of §3.1.4.2 is that the truncated identifiers specifically are rejected. As written, the test stays green if a future change makes this fixture fail for an unrelated reason such as a missing required property, and it would then be asserting only "parse failed somehow" while still reporting the requirement as verified.

Matching the structured error variant pins it to the actual cause. #315 built the `ModelError`-through-serde machinery precisely so tests would not have to string-match `Display` output, so the typed path is already available here.

crates/rstix/tests/interop/use_cases/attack_pattern/producer.rs
```diff
@@ -0,0 +1,320 @@
...
+    let (_, object_id) = load_attack_pattern(FIXTURE_CREATE);
+    let wire = objects
+        .iter()
+        .find(|obj| obj.get("id").and_then(Value::as_str) == Some(object_id.as_str()))
+        .expect("wire attack-pattern");
+    let created = wire
+        .get("created")
+        .and_then(Value::as_str)
+        .expect("created timestamp");
+    assert!(
```
mostafa: Minor: `contains('.') && ends_with('Z')` is a loose proxy for the millisecond precision the doc comment claims. It accepts `...T00:00:00.1Z` (decisecond) and `...T00:00:00.123456Z` (microsecond) just as readily as three digits.

If the intent is the interop millisecond requirement, assert exactly three subsecond digits. P-12 has the same check for `modified`, so both move together.

crates/rstix/tests/interop/use_cases/attack_pattern/consumer.rs
```diff
@@ -0,0 +1,217 @@
...
+    );
+    assert!(
+        bundle
+            .get_typed::<Vulnerability>(&relationship.target_ref)
+            .is_some()
+    );
+}
+
+/// REQ-CHK-SXC-3.1 / §4.2 Table 55 — Consumer handles §3.1.3 Producer test case data.
+pub fn assert_handles_producer_testcases() {
```
mostafa, line 212: This is the consumer half of the checklist duplication noted on `producer.rs`. The body is identical to `assert_producer_testcase_data` apart from the panic message: both iterate `PRODUCER_FIXTURES` and call `validate_interop_fixture`.

Re-validating producer data proves the producer requirement again rather than proving anything about consumption. Something consumer-shaped, such as asserting each fixture's use-case objects resolve to typed SDOs and their references close, would distinguish `REQ-CHK-SXC-3.1` from its SXP counterpart. The helpers you already have in `common/` would cover it without much new code.

mostafa (approved):

PR #414, opened 2026-07-30
fix(rstix): reproduce STIX id parse errors on serde recovery

mostafa:
## Summary

Follow-up to #413. `ParseError::InvalidStixId` recovery rebuilt `StixIdError::InvalidUuid` by re-parsing a hardcoded stand-in id and discarding the detail it had just serialized, so the recovered error described the stand-in rather than the id that was rejected. Because `validate::parse_bridge` builds the `STIX-E0003` diagnostic message from that error, the diagnostic misreported the defect: an id whose final UUID group is one character short was reported as though it had the wrong number of groups.

Three problems, all in `crates/rstix/src/serde_impls/stix_id.rs`:

- The recovered `InvalidUuid` carried the stand-in's `uuid` crate detail instead of the input's, which is the opposite of the fidelity the tagging was added for. The existing test only asserted `matches!(.., InvalidUuid(_))`, so it could not see this.
- The stand-in arm called `panic!` if `StixId::parse("x--not-a-uuid")` ever stopped yielding `InvalidUuid`. That sits on a path reachable from parsing untrusted documents, which the parse fuzz targets exist to keep panic-free.
- Rebuilding `StixIdError::TypeMismatch` required `Box::leak` to produce the variant's `&'static str`, leaking the decoded type name on every recovery. The arm was also unreachable: `StixId::parse` only returns `MissingDelimiter`, `EmptyTypeName`, or `InvalidUuid`, so tagged recovery never carried `TypeMismatch` in the first place.

## Approach

The tagged payload now carries the rejected id, and recovery replays `StixId::parse` on it, so the recovered error is the original by construction for every variant that parse can produce, `uuid` detail included. That removes the wire enum, the leak, the panic, and the unreachable arm, and leaves the module about 20 lines shorter.

Two properties preserved from the original design:

- Injection resistance. The payload is JSON encoded, so the U+001E tag cannot appear inside it, and decoding still requires the tag at position 0.
- Graceful degradation. A malformed, forged, or untagged message yields `None` and falls back to `ParseError::Json` rather than panicking.

Ids longer than 256 bytes are left untagged, so a hostile document cannot inflate an error message by the length of its own input. A STIX id is `{type}--{uuid}`, so real ids stay far below that.

Tests cover exact reproduction across all three parse error variants, detail fidelity against the stand-in that used to be substituted, the length cap, and non-recovery for forged or valid-id payloads.

Also in this PR: the #413 changelog entry listed requirement ids as a review response and used rustdoc link syntax that renders as literal brackets in release notes, and the P-05 wire type check left behind a no-op rebinding.

## Test plan

- [x] `cargo clippy -p rstix --no-default-features --all-targets --locked -- -D warnings`
- [x] `cargo doc --workspace --all-features --locked --no-deps`

No `docs/` content change: the site does not embed `CHANGELOG.md`, and `docs/content/library/rstix.md` names `ParseError` without enumerating variants, so nothing there goes stale.

## Typed references

@SecurityEnthusiast caught a factual error in the original description: typed-id conversion does not cross the serde boundary as `ModelError::IdTypeMismatch`. `define_typed_id!` mapped `from_stix_id` failures through a bare `serde::de::Error::custom`, and `ModelError::IdTypeMismatch` is the unrelated `validate_id_matches_type` check comparing an object's id prefix against its declared `type`.

Rather than restate the gap, this PR closes it. `created_by_ref` is `Option<IdentityId>` on the common SDO/SRO properties, so the path is reachable from any ordinary bundle: a `created_by_ref` holding a `malware--` id produced `ParseError::Json` and a generic `STIX-E0001`, saying nothing about the type the property requires.

Typed-ID conversion failures are now tagged with the expected and found type names, which is everything `StixIdError::TypeMismatch` carries, and recovery rebuilds the variant. Parsing that bundle now yields `ParseError::InvalidStixId(StixIdError::TypeMismatch { expected: "identity", found: "malware" })`, and the pipeline reports `STIX-E0021`, wrong target type for a typed reference, with a suggestion naming the required type. Malformed-id variants keep `STIX-E0003` and the `id` property path.

`expected` is a `&'static str`, and reconstructing one is exactly what the `Box::leak` this PR removes was for. It is not needed: the expected type is a `define_typed_id!` literal, so `StixObjectKind` already owns the identical `&'static str` and recovery looks it up, which also means a name the registry does not recognize yields no recovery instead of a fabricated error. `StixObjectKind::as_str` is public for that lookup. `StixIdError` is unchanged, so nothing breaks for consumers.

Removing the `TypeMismatch` arm from the `StixId` tagging is unaffected by any of this: `StixId::parse` cannot produce that variant, so that arm was unreachable regardless.


mostafa:
@SecurityEnthusiast Please review. 🙏

SecurityEnthusiast:
> @SecurityEnthusiast Please review. 🙏

Thanks @mostafa 

This is the right fix for the #413 recovery bug. Tagging the rejected id and replaying `StixId::parse` restores the original error (including `uuid` detail) by construction, drops the placeholder / panic / `Box::leak` path, and the new tests actually pin detail fidelity instead of only matching `InvalidUuid(_)`. Length cap + forged/valid-id non-recovery correct. Changelog and P-05 cleanup are fine.

One factual correction on the PR body:

Typed-id conversion does **not** currently cross the serde boundary as `ModelError::IdTypeMismatch`. On this branch it is still:

```rust
Self::from_stix_id(id).map_err(serde::de::Error::custom)
```

That is untagged `StixIdError` Display (typically `ParseError::Json`). `ModelError::IdTypeMismatch` is the separate object `type` / `id`-prefix check in model validate.

Dropping `TypeMismatch` from **StixId-tagged** recovery is still correct — `StixId::parse` cannot produce it.

I can just add 1 comment that 

1. Mention it that typed-id `TypeMismatch` remains Display/`ParseError::Json` for now, or  
2. leave an explicit deferred note if structured typed-id recovery is intentional follow-up.

mostafa:
You are right, and thanks for the precise correction. I checked both paths:

- `define_typed_id!` generates `Self::from_stix_id(id).map_err(serde::de::Error::custom)`, so a typed-id `TypeMismatch` crosses as an untagged Display string and lands in `ParseError::Json`.
- `ModelError::IdTypeMismatch` is produced by `validate_id_matches_type`, comparing an object's id prefix against its declared `type`. Different check, different data, different call site.

So my justification was wrong even though the removal is right for the reason you state: `StixId::parse` cannot produce `TypeMismatch`, so the tagged path never carried it and the `Box::leak` served nothing.

Corrected in `02e8d54` and in the PR description, which now says tagged recovery never carried the variant and states plainly that typed-id conversion reports `TypeMismatch` through `serde::de::Error::custom` and surfaces as `ParseError::Json`.

On your offer of a note: a plain statement of current behaviour is useful, so that is what both the changelog entry and the description now carry. I would rather not word it as "for now" or as a deferred item, since this repo keeps forward-looking notes out of the changelog and code comments. If structured recovery for typed-id conversion is worth having, it is its own change with a real design question attached, because `StixIdError::TypeMismatch::expected` is a `&'static str` and reconstructing one after decoding either needs a lookup over the known type names or a signature change to that field. Leaking to satisfy the lifetime is the thing this PR removes, so I did not want to reintroduce it under a different name here.

mostafa:
Closed the gap rather than noting it, since `created_by_ref` is `Option<IdentityId>` on the common SDO/SRO properties, so this is reachable from any ordinary bundle and not just from library callers holding typed ids directly. Before, a `created_by_ref` pointing at a `malware--` id gave `ParseError::Json` and a generic `STIX-E0001` that said nothing about the type the property requires.

`1d6f8dd` tags typed-ID conversion failures with the expected and found type names, which is everything `StixIdError::TypeMismatch` carries, and recovery rebuilds the variant. That bundle now yields `ParseError::InvalidStixId(StixIdError::TypeMismatch { expected: "identity", found: "malware" })`, and the pipeline reports `STIX-E0021`, wrong target type for a typed reference, with a suggestion naming the required type. Malformed-id variants keep `STIX-E0003` and the `id` property path.

The `&'static str` on `expected` is the thing that made the old reconstruction reach for `Box::leak`, and it turns out not to need an allocation at all: the expected type is a `define_typed_id!` literal, so `StixObjectKind` already owns the identical `&'static str` and recovery looks it up. A consequence worth having is that a type name the registry does not recognize yields no recovery instead of a fabricated error, which is the same fail-soft posture as the forged-payload path. `StixObjectKind::as_str` is public for the lookup and `StixIdError` is untouched, so there is no breaking change.

Tests: exact round-trip of the mismatch through the serde boundary, recovery from the deserializer, resolution across all four object classes the macro covers, the length cap on the found type name, forged payloads including a self-consistent `["identity","identity"]`, plus bundle-level `ParseError` and pipeline `STIX-E0021` assertions on the message and the fix suggestion.

Full gate set green, including `--no-default-features` clippy and `+1.88.0`. The description now carries this instead of the wrong claim.

SecurityEnthusiast:
Thanks @mostafa, that closes it properly and help to make next steps more consistent.

Replaying `StixId::parse` on the tagged rejected id was already the right StixId recovery. Closing typed-id the same way, without a leak or an `expected: String` break.
`created_by_ref: Option<IdentityId>` makes the path ordinary bundle traffic, and `ParseError::InvalidStixId(TypeMismatch { expected: "identity", found: "malware" })` + `STIX-E0021` is what that input should produce.

Looking up the expected name through `StixObjectKind` is the clean answer to the `&'static str` problem. Fail-soft when the name is not in the registry matches the forged-payload posture. Tests covering round-trip, deserializer recovery, all four object classes, the found-name length cap, self-consistent forged pairs, and the bundle/`STIX-E0021` path are what it should be.

PR #418, opened 2026-07-30
test(rstix): tighten §3.1 EX-4.2 / P-04 / C-04 evidence after #413/#414

SecurityEnthusiast:
## Summary

Follow-up to #413/#414: test-only soft-gap closes (no StixId/serde changes). EX-4.2 pins truncated wire ids and requires parse error equality; P-04 scopes Zero diagnostics by object_id only; C-04 matches query/typed fields to wire cardinality instead of overlapping C-01.

## Test plan

- [x] cargo fmt --all -- --check
- [x] cargo clippy -p rstix --all-targets --features validate,marking,graph --locked -- -D warnings
- [x] cargo test -p rstix --features validate,marking,graph --locked --test interop -- use_cases::attack_pattern
- [x] cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
- [x] cargo test --workspace --all-features --locked
- [x] cargo +1.88.0 check --workspace --all-features --locked


crates/rstix/tests/interop/use_cases/attack_pattern/examples.rs
```diff
@@ -43,26 +45,57 @@ pub fn assert_framework_example_rejects_truncated_ids() {
...
+        Err(err @ StixIdError::InvalidUuid(_)) => err,
+        other => panic!("truncated malware id must be InvalidUuid, got {other:?}"),
+    };
+    let relationship_err = match StixId::parse(TRUNCATED_RELATIONSHIP_ID) {
+        Err(err @ StixIdError::InvalidUuid(_)) => err,
+        other => panic!("truncated relationship id must be InvalidUuid, got {other:?}"),
+    };
+    match (&malware_err, &relationship_err) {
+        (StixIdError::InvalidUuid(_), StixIdError::InvalidUuid(_)) => {}
+        _ => unreachable!(),
```
mostafa: This block cannot fail. Both bindings come from `match` arms directly above that already panic on anything other than `InvalidUuid`, so re-matching the pair re-checks a guarantee that holds by construction, and the `unreachable!()` is a panic path with no reachable input. It reads as if it exists only to give `relationship_err` a use.

Suggest dropping it and asserting the relationship id without binding it, which leaves `malware_err` for the equality check below:

```rust
assert!(
    matches!(
        StixId::parse(TRUNCATED_RELATIONSHIP_ID),
        Err(StixIdError::InvalidUuid(_))
    ),
    "truncated relationship id must be InvalidUuid"
);
```

crates/rstix/tests/interop/use_cases/attack_pattern/examples.rs
```diff
@@ -43,26 +45,57 @@ pub fn assert_framework_example_rejects_truncated_ids() {
...
 
     let err = Bundle::parse_with_options(&fixture.json, &ParseOptions::new().interop_bundle())
         .expect_err("§3.1.4.2 truncated UUID ids must fail parse");
     match err {
-        ParseError::InvalidStixId(StixIdError::InvalidUuid(_)) => {}
-        other => panic!("expected ParseError::InvalidStixId(InvalidUuid), got: {other:?}"),
+        ParseError::InvalidStixId(got) => {
+            assert_eq!(
+                got, malware_err,
+                "bundle must fail on objects[1] malware id (first truncated id in walk order)"
```
mostafa: This message claims more than the assertion can show. Both published ids truncate the final UUID group to 11 characters at the same offset, so their errors are identical values:

```
malware--1121ffbc-364f-857a-9987-92fbcff24ab      -> InvalidUuid(Error(ParseGroupLength { group: 4, len: 11, index: 25 }))
relationship--11220001-3940-0405-20ff-1029b0bc922 -> InvalidUuid(Error(ParseGroupLength { group: 4, len: 11, index: 25 }))
```

They compare equal, so `got == malware_err` holds whether parse failed on objects[1] or objects[2]. Walk order happens to be deterministic (rstix bundle parse is sequential; the parallel work in #415 is daemon-side), but this assertion is not what establishes it, and if the failure ever moved to objects[2] the test would keep passing while naming objects[1].

The equality is still worth keeping, because it pins the `uuid` detail rather than just the variant and that detail is exactly what #414 fixed. It is the wording that should change, something like: bundle parse fails with the same `InvalidUuid` detail the published truncated ids carry, which both share. If you want position pinned as well, that needs evidence the ids cannot give, since the fixture must keep the published values.

crates/rstix/tests/interop/use_cases/attack_pattern/consumer.rs
```diff
@@ -148,22 +159,40 @@ pub fn assert_processes_fields() {
...
-        "{relative}: kill_chain_phases required for leaf processing"
+    let wire_phase_len = wire
+        .get("kill_chain_phases")
+        .and_then(Value::as_array)
+        .map(Vec::len)
+        .unwrap_or(0);
+    assert_eq!(
+        ap.kill_chain_phases.len(),
+        wire_phase_len,
+        "{relative}: typed kill_chain_phases length must match wire"
```
mostafa: This swaps C-04's own evidence for evidence C-01 already provides, which is the opposite of the intent.

`FIXTURE_TARGETS` is in `PRODUCER_FIXTURES`, so C-01 runs over this same fixture and object and already asserts `!ap.kill_chain_phases.is_empty()`, `!ap.common.external_references.is_empty()`, and `assert_wire_object_preserved`. That last one serializes the parsed object and requires wire containment element by element, which is strictly stronger than comparing lengths. So the cardinality checks here add no coverage C-01 lacks.

Meanwhile the `!is_empty()` assertions they replace were what guaranteed C-04's distinguishing evidence, the `KillChainPhase::validate` / `ExternalReference::validate` loops, actually executes. As written, a fixture that lost these arrays would satisfy `0 == 0`, skip both loops, and still pass while the doc comment claims leaf validation. Today that is caught only indirectly by C-01 asserting non-empty elsewhere.

The fixture has exactly one of each, so pinning the count keeps the cardinality intent and restores the guarantee in one line:

```rust
assert_eq!(ap.kill_chain_phases.len(), 1, "{relative}: one kill_chain_phase for leaf processing");
```

Same for `external_references` at line 195.

mostafa (changes requested):
P-04 is right and I would take it as-is. The `d.message.contains(object_id)` branch was the only Display-matching left in the function, and dropping it costs nothing because the same test already asserts `report.is_valid()` under `interop_bundle_strict`, which is Zero leniency over the whole bundle and therefore strictly subsumes the scoped filter. The `get_field` wire equalities in C-04 are also a real addition.

Three things on the other two, plus a rebase. The first is the one that matters, because it is the same over-claiming this PR sets out to remove.

I verified the EX-4.2 premise locally rather than reading it off the diff: both published truncated ids produce byte-identical errors, `InvalidUuid(Error(ParseGroupLength { group: 4, len: 11, index: 25 }))`, because both truncate the final group to 11 characters at the same offset. So the equality assertion cannot distinguish objects[1] from objects[2], which is what its message claims it establishes.

On the rebase: the branch conflicts with `main` in `CHANGELOG.md` only (the three test files merge clean). While you are there, `after #413 / #414 :` has a space before the colon, and the entry repeats the objects[1] claim, so it should say whatever the code ends up asserting.

One note on running this locally, since it cost me a minute: the harness is silent on success, so a clean run prints nothing at all. `target/interop-report/summary.json` is the confirmation it ran, and it reports 51/51 tested rows passing on this branch. `fmt` and `clippy` are green here too.

mostafa (commented):
All three are addressed, and EX-4.2 came back better than what I asked for. Rather than only rewording the claim, it now asserts the premise: `truncated_err` is pinned to the malware id's error, both published ids are required to share that detail, and the bundle-parse equality is stated as carrying the shared detail with the doc comment saying outright that it does not establish which object failed. That makes the shared-detail fact a test rather than a comment that can drift.

C-04 is what I hoped for too. `len() == 1` on both leaf lists keeps the cardinality intent while guaranteeing the `KillChainPhase` and `ExternalReference` validate loops are non-vacuous, and dropping the wire length computation removes the duplication of C-01's wire containment. The `wire` lookup still earns its place through the `name` and `created_by_ref` equalities. The tautological pair match and its `unreachable!()` are gone.

I verified rather than assumed: `fmt`, `clippy -p rstix --all-targets --features validate,marking,graph`, and the interop suite are green here, with 51/51 tested rows passing. I also mutated the published relationship id to check the new assertions actually run, and the suite failed at `examples.rs:72` with `objects[2].id must remain the published truncated relationship id`, so the row executes rather than being skipped. The rebase is clean and the changelog entry now says what the code asserts.

One leftover, inline below, and it is only whitespace in an entry this PR has nothing to do with. Happy to approve once that is back.

Optional, take it or leave it: the `matches!(StixId::parse(TRUNCATED_RELATIONSHIP_ID), Err(InvalidUuid(_)))` assert is now implied by the `assert_eq!` against `truncated_err` that follows, since that error is already pinned to `InvalidUuid`. It reads fine as explicit documentation, so I would not bother changing it.

CHANGELOG.md
```diff
@@ -29,7 +33,6 @@ The processor now reports failed input indices with its batch result. DLQ-enable
 On the pinned SigmaHQ raw Windows workload with `--logsource-routing`, `--batch-size 512`, and 16 k6 VUs, the median of three same-build runs moved from 306,351 to 384,906 events/s at eight threads while one-thread throughput remained near 90,000 events/s. Eight-core scaling improved from 3.39x to 4.19x, or from 42% to 52% efficiency, and the measured parse share fell from about 40% to 27%. A zero-copy request-body span prototype did not improve throughput and was not retained.
 
 `rsigma_batch_phase_duration_seconds{phase="parse"|"evaluate"}` exposes cumulative batch time for the two measured regions. `scripts/perf/baseline-daemon.sh` reports their deltas and parse share after each run.
-
 ### jemalloc as the musl global allocator (#412)
```
mostafa, line 37: Rebase artifact: the blank line between the #415 entry's last paragraph and this heading was dropped, so this heading now follows the paragraph directly. Every other section in the file keeps that blank line, and #412 is unrelated to this PR, so the diff should not touch it.

CommonMark still renders it as a heading, since an ATX heading can interrupt a paragraph, so nothing breaks in the published notes. It is just stray noise in someone else's entry. Restoring the blank line is the whole fix.

mostafa: I'll fix it.

mostafa (approved):
Restored the blank line myself in `e2e9425` so this does not need another round trip. The changelog diff against `main` is now purely additive, touching only the #418 entry.

Everything else stands as reviewed: EX-4.2 asserts the shared-detail premise rather than describing it, C-04 pins `len() == 1` so the leaf validate loops cannot go vacuous, and P-04 drops the last Display matching without losing coverage that `report.is_valid()` under Zero leniency does not already provide.

Verified locally before approving: `fmt`, `clippy -p rstix --all-targets --features validate,marking,graph`, and the interop suite green at 51/51 tested rows, plus a mutation of the published relationship id to confirm the new assertions actually execute. Approving pending CI on the final commit.

mostafa (approved):

PR #433, opened 2026-08-02
test(rstix): OASIS §3.2 Campaign Sharing interop use-case tests

SecurityEnthusiast:
## Summary

Adds §3.2 Campaign Sharing interop tests - CSD01 §3.2.1–§3.2.6. **2/21** use cases; not OASIS SXP/SXC certification.

- Description + Producer `P-01`..`P-11` + Consumer `C-01`..`C-05`
- Checklist `REQ-CHK-SXP-3.2` / `REQ-CHK-SXC-3.2`
- Non-gating examples `EX-4.1` / `EX-4.2`
- Table 4 typo (`type` = `threat-actor`) ignored; tests enforce `campaign`

## Test plan

- [*] `cargo test --workspace --all-features --locked`
- [*] `cargo +1.88.0 check --workspace --all-features --locked`
- [*] `cargo clippy -p rstix --all-targets --features validate,marking,graph --locked -- -D warnings`
- [*] `cargo test -p rstix --features validate,marking,graph --locked --test interop -- use_cases::campaign`
- [*] `summary.json` shows 72/72 TESTED passed

### DIFF
--- a/crates/rstix/tests/interop/use_cases/attack_pattern/producer.rs
+++ b/crates/rstix/tests/interop/use_cases/attack_pattern/producer.rs
@@ -180,15 +180,13 @@ pub fn assert_prop_spec_version() {
 pub fn assert_prop_id() {
     let fixture = load_fixture(FIXTURE_CREATE);
     let objects = parse_fixture_objects(&fixture.json).expect("parse fixture");
-    let (_, object_id) = load_attack_pattern(FIXTURE_CREATE);
-    let wire = objects
-        .iter()
-        .find(|obj| obj.get("id").and_then(Value::as_str) == Some(object_id.as_str()))
-        .expect("wire attack-pattern");
-    let wire_id = wire
-        .get("id")
-        .and_then(Value::as_str)
-        .expect("wire attack-pattern id");
+    let use_case_ids = use_case_object_ids(FIXTURE_CREATE, &objects);
+    assert_eq!(
+        use_case_ids.len(),
+        1,
+        "{FIXTURE_CREATE}: expected one attack-pattern use-case id"
+    );
+    let wire_id = &use_case_ids[0];
     assert!(
         wire_id.starts_with("attack-pattern--"),
         "id must use attack-pattern-- prefix: {wire_id}"
--- a/crates/rstix/tests/interop/use_cases/attack_pattern/producer.rs
+++ b/crates/rstix/tests/interop/use_cases/attack_pattern/producer.rs
@@ -9,32 +9,12 @@ use serde_json::Value;
 
 use crate::common::fixture_catalog::{parse_fixture_objects, use_case_object_ids};
 use crate::common::identity::assert_identity_shape;
+use crate::common::timestamp::assert_millisecond_rfc3339;
 use crate::harness::fixture::load_fixture;
 use crate::harness::interop_gate::{
     InteropGateOptions, validate_interop_fixture, validate_interop_json,
 };
 use crate::use_cases::attack_pattern::{FIXTURE_CREATE, PRODUCER_FIXTURES};
-
-/// Interop millisecond timestamps: exactly three fractional digits before `Z`.
-fn assert_millisecond_rfc3339(label: &str, value: &str) {
-    let Some((_, frac_and_z)) = value.rsplit_once('.') else {
-        panic!("{label} must include fractional seconds: {value}");
-    };
-    assert!(
-        frac_and_z.ends_with('Z'),
-        "{label} must end with Z: {value}"
-    );
-    let digits = &frac_and_z[..frac_and_z.len() - 1];
-    assert_eq!(
-        digits.len(),
-        3,
-        "{label} must have exactly three subsecond digits: {value}"
-    );
-    assert!(
-        digits.chars().all(|c| c.is_ascii_digit()),
-        "{label} fractional part must be digits: {value}"
-    );
-}
 
 fn load_attack_pattern(relative: &str) -> (AttackPattern, String) {
     let fixture = load_fixture(relative);
@@ -71,20 +51,39 @@ pub fn assert_select_content() {
         .get_mut("objects")
         .and_then(Value::as_array_mut)
         .expect("objects array");
+    let mut renamed = 0usize;
     for object in objects.iter_mut() {
         if object.get("type").and_then(Value::as_str) == Some("attack-pattern") {
             object["name"] = Value::String("Caller-selected Attack Pattern name".into());
+            renamed += 1;
         }
     }
+    assert_eq!(
+        renamed, 1,
+        "expected exactly one attack-pattern object renamed by caller selection"
+    );
     let json = serde_json::to_string(&root).expect("serialize caller-selected bundle");
     let use_case_ids = use_case_object_ids(FIXTURE_CREATE, &parse_fixture_objects(&json).unwrap());
-    validate_interop_json(
+    let bundle = validate_interop_json(
         &json,
         &InteropGateOptions {
-            use_case_object_ids: use_case_ids,
+            use_case_object_ids: use_case_ids.clone(),
         },
     )
     .expect("caller-selected bundle must pass interop gate");
+    assert_eq!(
+        use_case_ids.len(),
+        1,
+        "caller-selected bundle must expose one attack-pattern use-case id"
+    );
+    let stix_id = StixId::parse(&use_case_ids[0]).expect("attack-pattern id");
+    let ap = bundle
+        .get_typed::<AttackPattern>(&stix_id)
+        .expect("typed attack-pattern after caller selection");
+    assert_eq!(
+        ap.name, "Caller-selected Attack Pattern name",
+        "caller-selected name must survive parse and re-validation"
+    );
 }
 
 /// REQ-3.1-P-03 — Identity in bundle complies with §2.3.4 (fixture-scoped; not a duplicate §2.3 proof).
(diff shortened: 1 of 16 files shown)


crates/rstix/tests/interop/use_cases/campaign/producer.rs
```diff
@@ -0,0 +1,366 @@
...
+    }
+    let json = serde_json::to_string(&root).expect("serialize caller-selected bundle");
+    let use_case_ids = use_case_object_ids(FIXTURE_CREATE, &parse_fixture_objects(&json).unwrap());
+    validate_interop_json(
+        &json,
+        &InteropGateOptions {
+            use_case_object_ids: use_case_ids,
+        },
+    )
+    .expect("caller-selected bundle must pass interop gate");
```
mostafa, line 77: P-02 passes even when the selection step matches nothing. I flipped `Some("campaign")` to a type that appears in no fixture and the row still recorded Pass, because the only assertion is that the (now unmodified) bundle clears the gate, which is what P-01 already proves.

Suggest pinning that the caller-selected content actually survived the round trip: count the objects renamed and assert it is 1, then assert the parsed bundle's typed Campaign carries `Caller-selected Campaign name`. That makes the row evidence that a caller-selected object set parses and re-validates, rather than evidence that the fixture parses.

The same shape is in `use_cases::attack_pattern::producer::assert_select_content`, so §3.1 has the identical gap and is worth fixing in the same pass.

crates/rstix/tests/interop/use_cases/campaign/description.rs
```diff
@@ -0,0 +1,50 @@
...
+    let attributed = load_fixture(FIXTURE_ATTRIBUTED);
+    let attributed_bundle = validate_interop_fixture(FIXTURE_ATTRIBUTED, &attributed.json)
+        .expect("§3.2.3.2 must parse for description-scope check");
+    assert_eq!(
+        attributed_bundle.objects_of_type::<Campaign>().count(),
+        1,
+        "§3.2.3.2 must carry a Campaign SDO as described in §3.2.1"
+    );
+}
+
```
mostafa, line 47: The doc comment says this check binds the description to "§3.2.3.2 carrying an Intrusion Set attribution path", but the only assertion made against that fixture is the Campaign count, so the attribution half of the claim is unbacked.

Adding `assert_eq!(attributed_bundle.objects_of_type::<IntrusionSet>().count(), 1, ...)` covers it and stays distinct from C-05, which resolves the relationship endpoints rather than asserting the SDO is present. Otherwise drop that clause from the doc so the prose matches the evidence.

crates/rstix/tests/interop/use_cases/campaign/producer.rs
```diff
@@ -0,0 +1,366 @@
...
+    assert_eq!(
+        digits.len(),
+        3,
+        "{label} must have exactly three subsecond digits: {value}"
+    );
+    assert!(
+        digits.chars().all(|c| c.is_ascii_digit()),
+        "{label} fractional part must be digits: {value}"
+    );
+}
```
mostafa: This helper is byte-identical to the one in `use_cases::attack_pattern::producer`, all 19 lines. The millisecond rule is not use-case specific, so it belongs in `common/` next to `fixture_catalog`, `identity`, and `wire_preservation`, where a single definition keeps the interpretation of "exactly three subsecond digits" from drifting between copies.

crates/rstix/tests/interop/use_cases/campaign/producer.rs
```diff
@@ -0,0 +1,366 @@
...
+pub fn assert_prop_id() {
+    let (_, object_id) = load_campaign(FIXTURE_CREATE);
+    assert!(
+        object_id.starts_with("campaign--"),
+        "id must use campaign-- prefix: {object_id}"
+    );
+    assert!(
+        StixId::parse(&object_id).is_ok(),
+        "id must be valid STIX id"
+    );
```
mostafa: Nit: this assertion cannot fail, since `load_campaign` already calls `StixId::parse(&object_id).expect("campaign id")` on the same string a few lines earlier. The `campaign--` prefix check above it is the part carrying the requirement, so either drop this or have P-07 read the id straight off the wire object and parse it independently.

mostafa (commented):
Reviewed against the §3.1 baseline. This carries over the hardening from #413/#414/#418 correctly: P-04 is genuinely distinct from P-01 and scopes Zero-failing diagnostics through the structured `object_id` rather than message matching, P-05 asserts the wire `type` (with the Table 4 `threat-actor` typo called out and `campaign` enforced), P-09/P-10 use the exact three-digit millisecond check, C-04 compares `get_field` results against the wire object, and REQ-CHK-SXC-3.2 closes `created_by_ref` to a typed Identity instead of re-running the gate.

Verified locally on the head commit:

- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, and `cargo +1.88.0 check --workspace --all-targets --all-features --locked` all clean. Clippy also clean on the pinned 1.88.0 toolchain for `-p rstix --all-targets --features validate,marking,graph`, which is where the `uninlined_format_args` lint slipped past CI before.
- `summary.json` reports 72 TESTED rows, 72 passed, 7 report-only, 2 blocked, 0 smoke. The 21 new manifest rows are all TESTED and every one has a registered `interop_test!`, and the README arithmetic (9 + 18 + 24 + 21) matches.
- Fixture provenance holds up: the missing opening quote on the §3.2.4.2 relationship id key is a MECHANICAL repair with defect 20 recorded, and the smart-quotes repair on §3.2.4.1 without a divergence entry matches the existing §3.1 example sidecars, so that asymmetry is pre-existing practice rather than something new here.

One row does not establish what it claims (P-02, verified by mutation), one doc comment claims more than its assertions cover, plus two smaller items inline.

mostafa (commented):
All four items addressed, and the §3.1 counterparts were fixed in the same pass, which is the right call since both use cases land in the same unreleased section.

Verified on the new head:

- P-02 is now live evidence in both use cases. I re-ran the mutation that previously passed silently: pointing the selection filter at a type no fixture contains now fails on `expected exactly one campaign object renamed by caller selection`, and changing the injected name so it no longer matches the expectation fails on `caller-selected name must survive parse and re-validation`. The row now proves a caller-selected object set round-trips rather than restating P-01.
- §3.2.1 now asserts the Intrusion Set SDO on the §3.2.3.2 fixture, so the prose about the attribution path is backed, and it stays distinct from C-05, which resolves the relationship endpoints.
- The millisecond helper moved to `common/timestamp.rs` with both producers importing it, so there is a single definition of the three-digit rule.
- `summary.json` still reports 72 TESTED rows, 72 passed. `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, and clippy on the pinned 1.88.0 toolchain for `-p rstix --all-targets --features validate,marking,graph` are all clean.

One cosmetic leftover on P-07 inline, not worth holding the PR for. Optionally, since #418's entry frames itself as closing the §3.1 soft gaps, a clause in the #433 entry noting that it also tightens §3.1 P-02 and P-07 would keep the release notes self-consistent.

crates/rstix/tests/interop/use_cases/campaign/producer.rs
```diff
@@ -0,0 +1,372 @@
...
+        .expect("wire campaign");
+    let wire_id = wire
+        .get("id")
+        .and_then(Value::as_str)
+        .expect("wire campaign id");
+    assert!(
+        wire_id.starts_with("campaign--"),
+        "id must use campaign-- prefix: {wire_id}"
+    );
+    StixId::parse(wire_id).expect("wire id must be valid STIX id");
```
mostafa, line 199: The parse here still cannot fail. `wire` is located by matching `obj["id"]` against `object_id`, and `object_id` was already run through `StixId::parse(...).expect(...)` inside `load_campaign`, so `wire_id` is that same validated string by construction.

Locating the object by `type` instead of by id would make the parse genuinely independent of the helper. Cosmetic either way, since the `campaign--` prefix check is what carries the requirement.

crates/rstix/tests/interop/use_cases/attack_pattern/producer.rs
```diff
@@ -179,15 +178,22 @@ pub fn assert_prop_spec_version() {
...
-        object_id.starts_with("attack-pattern--"),
-        "id must use attack-pattern-- prefix: {object_id}"
-    );
-    assert!(
-        StixId::parse(&object_id).is_ok(),
-        "id must be valid STIX id"
+        wire_id.starts_with("attack-pattern--"),
+        "id must use attack-pattern-- prefix: {wire_id}"
     );
+    StixId::parse(wire_id).expect("wire id must be valid STIX id");
```
mostafa: The parse here still cannot fail. `wire` is located by matching `obj["id"]` against `object_id`, and `object_id` was already run through `StixId::parse(...).expect(...)` inside `load_attack_pattern`, so `wire_id` is that same validated string by construction.

Locating the object by `type` instead of by id would make the parse genuinely independent of the helper. Cosmetic either way, since the `attack-pattern--` prefix check is what carries the requirement.

mostafa (commented):
Two cosmetic follow-ups on P-07, non-blocking.

mostafa (approved):

mostafa:
Pushed 2b42d78 to this branch rather than leave the cosmetic P-07 item open.

`assert_prop_id` now selects the wire object through `use_case_object_ids`, which returns raw wire id strings filtered by use-case type and never parses them, so `StixId::parse(wire_id)` is a first parse of an unvalidated string instead of restating the parse `load_campaign` and `load_attack_pattern` already performed. Both use cases changed the same way, and each asserts exactly one use-case id so the row cannot pass on an empty selection.

I also added a bullet to the #433 changelog entry recording the §3.1 P-02 and P-07 tightening and the shared timestamp helper move, since those land in the same release.

Verified on the pushed head: mutating the expected id prefix fails the row, `summary.json` still reports 72 TESTED rows with 72 passed, and fmt, workspace clippy with all features, clippy and check on the pinned 1.88.0 toolchain, and the `rstix` test suite are all clean.

PR #438, opened 2026-08-04
test(rstix): OASIS §3.3–§3.21 interop use-case tests

SecurityEnthusiast:
## Summary

- Adds OASIS STIX 2.1 Interoperability CSD01 **§3.3–§3.21** use-case modules under `crates/rstix/tests/interop/use_cases/` (Confidence through Vulnerability, including Versioning). This branch completes the remaining nineteen sections so all **21/21** use cases have dedicated modules.
- Manifest: **506** `TESTED` rows (9 harness + 18 §2.3 + 479 §3.x use-case rows). **`REQ-CHK-SXP-3.12`** and **`REQ-CHK-SXP-3.16`** stay **`BLOCKED`** on published defects 19 and 16. Five §4.2 checklist/framework rows remain `REPORT_ONLY`.
- Vendors non-gating example fixtures under `examples/` where CSD01 defines §3.x.4 / §3.x.7 content; records provenance repairs (bundle wrap, smart quotes, malformed ids) with `defect` where required.
- Small model helpers where consumer C-04 needed them: `QueryableStixObject::get_field` now answers `["created_by_ref"]` on Infrastructure, IntrusionSet, Location, Malware, MalwareAnalysis, Report, and Vulnerability (seven SDOs; completes coverage across all 19 SDOs that implement the trait). IntrusionSet additionally answers `description`, `resource_level`, and `primary_motivation`.
- Follow-up commit applies §3.1/§3.2 review lessons to new slices: versioning mod/rev P-02 now mutates + counts renames + asserts typed field survival; Note/Sighting/Versioning description rows assert the related typed objects their prose claims. No further changes to §3.1 or §3.2.

This is **not** OASIS SXP/SXC certification. Counts are harness `TESTED`/`Pass` rows against normative fixtures and examples.

## Test plan

- [x] `cargo clippy -p rstix --all-targets --features validate,marking,graph --locked -- -D warnings`
- [x] `cargo test -p rstix --all-features --locked`
- [x] Confirm `target/interop-report/summary.json`: `tested_rows_passed` == `manifest_rows_by_disposition.tested` == **506**, `blocked` == **2**
- [x] Spot-check: versioning mod/rev P-02 fails if the type filter matches nothing; Note description fails without Threat Actor; Sighting description fails if `sighting_of_ref` Indicator is missing


mostafa (commented):
Reviewed structurally plus sampled, rather than line by line: at 19 sections and roughly 23k lines this is about ten times the size of the §3.1 and §3.2 slices, and per-section scrutiny at the depth those got is not something a single pass over this can honestly claim. Splitting future work back into per-section PRs would keep the review depth matched to what the certification evidence is being asked to support.

What I verified mechanically, all of it clean:

- All 479 use-case rows resolve to a distinct assertion function, with no row registered against a shared or missing body, and no assertion comparing two literals. Manifest totals reconcile: 513 requirements as 506 TESTED plus 2 BLOCKED plus 5 REPORT_ONLY, and 506 equals 9 harness plus 18 §2.3 plus 479 use-case rows.
- The P-02 fix from #433 was carried into every producer module. All 23 mutate a type-appropriate property, count the mutations and assert exactly one, then assert the mutated value survives parse and re-validation through the typed object. The four that pick a non-name property (`region`, `number_observed`, `opinion`, `count`) are equally sound.
- Description rows separate the OASIS prose from what they bind. I checked the doc claims against the assertions across all 21 and found no repeat of the §3.2 over-claim.
- I probed for the vacuity risk in the `use_case_object_ids` loops by making that helper panic on an empty result. No call site returns empty, so the C-01 loops are all doing work today. Worth noting the latent shape: `use_case_types_for_fixture` falls through to an empty slice and the C-01 loops have no non-empty guard, so a fixture directory added without a mapping entry would silently empty those rows rather than fail.
- Both BLOCKED rows carry `blocked_by` with matching `divergence_recorded` entries, and the `published` typo behind defect 16 is handled coherently: unrepaired and blocked in the normative test case, repaired under `JUDGMENT` authority with a recorded divergence in the non-gating examples.
- `get_field` now answers `created_by_ref` on all 19 SDOs that implement `QueryableStixObject`, so the query surface ends up uniform rather than partial. No module unit-tests its `get_field` arms, which matches existing practice in the crate.
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, `cargo +1.88.0 check --workspace --all-targets --all-features --locked`, clippy on the pinned 1.88.0 toolchain for the lean rstix feature set, and `cargo test -p rstix --all-features --locked` all pass. `summary.json` reports 506 TESTED with 506 passed and 2 blocked.

One change needs to come out, plus a release-notes gap and a wording nit inline.

crates/rstix/tests/fixtures/interop/testcases/observed-data/tc-3.14.3.2-observed-data-domain-name-and-ip-address.provenance.toml
```diff
@@ -22,3 +22,8 @@ description = "Correct observed-data object_refs ipv4-address-- token to ipv4-ad
 authority = "MECHANICAL"
 class = "no-container"
 description = "Wrap objects in a bundle per REQ-2.3-X-02."
+
+[[repair]]
+authority = "MECHANICAL"
+class = "timestamp-normalization"
+description = "Normalize first_observed and last_observed to millisecond RFC 3339 for load-time scan parity."
```
mostafa: This rewrites published normative data for a reason that does not hold, and the edit is not needed.

The load-time scan checks millisecond precision on `created`, `modified`, `published`, `first_seen`, and `last_seen` only. It has never covered `first_observed`, `last_observed`, or `valid_from`, and this PR does not change that list, so there is no scan to bring these properties into parity with. STIX 2.1 sets a millisecond minimum only for `created` and `modified`, which means `2015-12-21T19:00:00Z` as published was already spec-legal.

I restored both this fixture and the opinion one to their published values and ran `cargo test -p rstix --all-features --locked`: everything passes and `summary.json` still reports 506 TESTED with 506 passed. So the normalization buys nothing while changing the material the §3.14 and §3.15 evidence is measured against, which is the one category of fixture edit worth avoiding.

Please revert the two JSON edits and drop these two `[[repair]]` entries. The same question applies to the three example sidecars that normalize `valid_from`, `first_observed`, or `last_observed` for keys the scan does not read. If the intent is that the harness should require millisecond precision on those properties, that belongs in the scan key list with the requirement stated there, not in pre-normalized fixtures.

crates/rstix/tests/fixtures/interop/testcases/opinion/tc-3.15.3.1-opinion-on-indicator-different-identity.provenance.toml
```diff
@@ -12,3 +12,8 @@ description = "Unescape spec field names on identity, indicator, and opinion obj
 authority = "MECHANICAL"
 class = "no-container"
 description = "Wrap objects in a bundle per REQ-2.3-X-02."
+
+[[repair]]
+authority = "MECHANICAL"
+class = "timestamp-normalization"
+description = "Normalize indicator valid_from to millisecond RFC 3339 for load-time scan parity."
```
mostafa: Same as the observed-data sidecar: `valid_from` is not one of the scanned keys, the published `2016-01-01T00:00:00Z` is spec-legal, and the suite is green at 506/506 with the published value restored. Revert the JSON edit and drop this entry.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to RSigma are documented in this file. Each entry correspond
 
 ## [Unreleased]
 
+### rstix: OASIS §3.3–§3.21 interop use-case tests (#438)
```
mostafa, line 7: The entry describes only the test additions, but this PR also changes public library behavior: `QueryableStixObject::get_field` now answers `["created_by_ref"]` on Infrastructure, IntrusionSet, Location, Malware, MalwareAnalysis, Report, and Vulnerability, and IntrusionSet additionally gained `description`, `resource_level`, and `primary_motivation`. That completes `created_by_ref` coverage across all 19 SDOs, which is a genuinely useful thing for callers to learn from the release notes.

Worth a sentence here. The PR body also describes this as five types when it is seven.

crates/rstix/tests/interop/use_cases/sighting/producer.rs
```diff
@@ -0,0 +1,440 @@
...
+        "caller-selected bundle must expose one sighting use-case id"
+    );
+    let stix_id = StixId::parse(&use_case_ids[0]).expect("sighting id");
+    let sighting = bundle
+        .get_typed::<Sighting>(&stix_id)
+        .expect("typed sighting after caller selection");
+    assert_eq!(
+        sighting.count,
+        Some(99),
+        "caller-selected name must survive parse and re-validation"
```
mostafa: Nit: the message says `name` but the row mutates `count`. Copy-paste from the campaign version.

crates/rstix/tests/fixtures/interop/testcases/observed-data/tc-3.14.3.2-observed-data-domain-name-and-ip-address.provenance.toml
SecurityEnthusiast: Fixed in `cbaafa3f`.

Restored the published `first_observed` / `last_observed` values (`2015-12-21T19:00:00Z`) and dropped the unjustified `timestamp-normalization` repair on this sidecar. Applied the same rule to the related example sidecars and to the parallel `valid_from` “scan parity” repair on `tc-3.16.3.1`.

Remaining `timestamp-normalization` entries are intentional and in-scope only:
- report examples: JUDGMENT repair of defective `published` (defect 16), which **is** a load-time scan key
- sighting example: MECHANICAL ms only for scanned `first_seen` / `last_seen`

crates/rstix/tests/fixtures/interop/testcases/opinion/tc-3.15.3.1-opinion-on-indicator-different-identity.provenance.toml
SecurityEnthusiast: Fixed in `cbaafa3f`.

Restored published `valid_from` (`2016-01-01T00:00:00Z`) and dropped the sidecar repair. Same treatment on the opinion example sidecar.

CHANGELOG.md
SecurityEnthusiast: Fixed in `cbaafa3f`.

CHANGELOG now documents that `QueryableStixObject::get_field` answers `["created_by_ref"]` on Infrastructure, IntrusionSet, Location, Malware, MalwareAnalysis, Report, and Vulnerability (seven SDOs; completes all 19), plus IntrusionSet `description` / `resource_level` / `primary_motivation`. PR body corrected from five to seven.

crates/rstix/tests/interop/use_cases/sighting/producer.rs
SecurityEnthusiast: Fixed in `cbaafa3f`: assert message now says `count`, matching the mutated property.

SecurityEnthusiast:
Follow-ups from the review notes (not the three required change items):

- **Remaining `timestamp-normalization` repairs** stay on purpose: report examples (defect 16 `published`, a scanned key) and the sighting example (ms only for scanned `first_seen`/`last_seen`). Unscanned key rewrites were reverted in `cbaafa3f`.
- **Per-section PRs**: agreed for future interop slices — review depth should match certification-evidence size.
- **Empty-mapping vacuity**: closed locally in `fixture_catalog.rs` — `use_case_types_for_fixture` now panics on unknown `testcases/<section>/`, and `use_case_object_ids` asserts a non-empty match so C-01 loops cannot silently do zero work. Interop suite still **506/506** after the guard.

mostafa (commented):
All addressed, and two of them beyond what I asked for.

- The published timestamps are back. `tc-3.14.3.2` and `tc-3.15.3.1` are now byte-identical to `main`, and the only normative fixture still differing is `tc-3.16.3.1`, where you also reverted the pre-existing `valid_from` normalization that predates this PR and dropped its repair entry. That is the right cleanup to fold in here.
- The example fixtures follow the same rule now: `valid_from`, `first_observed`, and `last_observed` carry their published second-precision values, and the one remaining `timestamp-normalization` on the sighting example is narrowed to `first_seen` and `last_seen`, which the load-time scan does read, with the description reworded to say so. The two report examples correctly keep their defect 16 `published` repair under `JUDGMENT` authority, since that one fixes genuinely malformed data.
- The changelog now states the query-surface change, naming the seven SDOs, the completion across all 19, and the extra IntrusionSet paths.
- Fail-closed mapping is a better answer than the note I left. `use_case_types_for_fixture` panics on an unmapped or non-testcases path and `use_case_object_ids` asserts a non-empty match, so a fixture directory added without a mapping entry now fails loudly instead of quietly emptying the C-01 loops. I confirmed no call site passes an `examples/` path into either helper, statically and by the all-features run.
- The sighting P-02 message says `count`.

Verified on the new head: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, clippy on the pinned 1.88.0 toolchain for the lean rstix feature set, and `cargo test -p rstix --all-features --locked` all pass. `summary.json` reports 506 TESTED with 506 passed, 2 blocked, 5 report-only, and all 479 use-case rows still resolve to distinct assertion functions.

No further comments from me.

mostafa (approved):

PR #440, opened 2026-08-04
test(rstix): STIX 2.1 interop self-certification CI gate

SecurityEnthusiast:
## Summary
- Adds a dedicated CI job that runs the full rstix interop suite (`cargo test -p rstix --test interop --features validate,marking,graph`) — every `TESTED` manifest row that backs CSD01 SXP/SXC **self-certification** (21 use cases; Tables 55/56 filled with Pass or BLOCKED).
- Stamps `generated_at` into `summary.json` and gates the run with `scripts/interop-report-gate.py` so a missing, stale, or incomplete report fails CI; uploads `target/interop-report/` (`summary.json`, `sxc-table-55.md`, `sxp-table-56.md`, `traceability.csv`, `risks.md`).
- This is operational self-certification against Interoperability **CSD01** (library as SXP + SXC). 
- `REQ-CHK-SXP-3.12` / `REQ-CHK-SXP-3.16` remain BLOCKED (defects 19 / 16).

## Test plan
- [*] `cargo +1.88.0 check --workspace --all-targets --all-features --locked`
- [*] `cargo clippy -p rstix --all-targets --features validate,marking,graph --locked -- -D warnings`
- [*] `cargo test -p rstix --test interop --features validate,marking,graph --locked`
- [*] `export INTEROP_RUN_START=$(date -u +%Y-%m-%dT%H:%M:%SZ) && cargo test -p rstix --test interop --features validate,marking,graph --locked && python3 scripts/interop-report-gate.py`
- [*] Confirm `target/interop-report/summary.json`: `tested_rows_passed` == 506, `blocked_rows` == 2, `generated_at` present
- [*] `cargo test -p rstix --test interop_sentinel --locked` fails without features; passes with `validate,marking,graph`

### DIFF
--- a/crates/rstix/README.md
+++ b/crates/rstix/README.md
@@ -951,13 +951,13 @@ cargo test -p rstix --test interop_sentinel --features validate,marking,graph --
 
 **CI:** the `rstix STIX interop self-certification` job in `.github/workflows/ci.yml` runs the full interop suite (every `TESTED` row), then `scripts/interop-report-gate.py` (report present, `generated_at` not older than `INTEROP_RUN_START`, Tables 55/56 cell content and traceability CSV rows match `gate-expectations.json`, every `TESTED` row passed, required features on), and uploads `target/interop-report/` as the `interop-report` artifact. The multi-OS `Test` matrix also executes the suite under `--all-features`; the dedicated job is the gated self-certification report producer.
 
-**Three-layer gate:**
+**Three-layer gate** (run order):
 
 | Layer | Where | What it catches |
 | ----- | ----- | ---------------- |
-| 2 — export invariants | `tests/interop/harness/certification.rs` | Empty/wrong checklist `Result`, incomplete CSV, export before coverage passes |
-| 3 — manifest sync | `tests/interop/harness/gate_expectations.rs` + committed `gate-expectations.json` | Stale golden expectations after `manifest.toml` changes |
-| 1 — CI artifact gate | `scripts/interop-report-gate.py` | Stale/missing report, wrong table cells, CSV row/order/outcome drift, summary count mismatch |
+| 1 — export invariants | `tests/interop/harness/certification.rs` | Empty/wrong checklist `Result`, incomplete CSV, export before coverage passes |
+| 2 — manifest sync | `tests/interop/harness/gate_expectations.rs` + committed `gate-expectations.json` | Stale golden expectations after `manifest.toml` changes |
+| 3 — CI artifact gate | `scripts/interop-report-gate.py` | Stale/missing report, wrong table cells, CSV row/order/outcome drift, summary count mismatch |
 
 Regenerate expectations after manifest edits:
 
--- a/crates/rstix/README.md
+++ b/crates/rstix/README.md
@@ -906,6 +906,7 @@ Test harness for **STIX 2.1 Interoperability Version 1.0, Committee Specificatio
 | Cross-cutting | `tests/interop/common/` | Automated checks for 18 §2.3 manifest rows, run suite-wide over **42** walkable normative fixtures |
 | Use-case tests | `tests/interop/use_cases/` | All 21 OASIS §3.x use cases (§3.1–§3.21) with Producer/Consumer/example modules — the SXP/SXC self-certification evidence |
 | Manifest | `tests/fixtures/interop/manifest.toml` | `req_id` → `test_id` → fixture → §4.2 checklist row |
+| Gate expectations | `tests/fixtures/interop/gate-expectations.json` | Manifest-derived golden rows for Tables 55/56 and traceability CSV (regenerate with `scripts/generate-interop-gate-expectations.py` when the manifest changes) |
 | Normative fixtures | `tests/fixtures/interop/testcases/` | Gating OASIS test-case data + `.provenance.toml` sidecars (**44/44** inventory: **42** suite-walkable + **2** `BLOCKED` on §9.1 defects 16 and 19; synthetic helpers remain for intentional negatives) |
 | Examples | `tests/fixtures/interop/examples/` | Non-normative; must never fail the build |
 | Report artifacts | `target/interop-report/` | Self-certification package: `summary.json` (incl. `generated_at`), Tables 55/56 markdown, traceability CSV, risks (never committed; CI uploads as `interop-report`) |
@@ -948,7 +949,21 @@ cargo test -p rstix --test interop_sentinel --locked
 cargo test -p rstix --test interop_sentinel --features validate,marking,graph --locked
 ```
 
-**CI:** the `rstix STIX interop self-certification` job in `.github/workflows/ci.yml` runs the full interop suite (every `TESTED` row), then `scripts/interop-report-gate.py` (report present, `generated_at` not older than `INTEROP_RUN_START`, every `TESTED` row passed, required features on), and uploads `target/interop-report/` as the `interop-report` artifact. The multi-OS `Test` matrix also executes the suite under `--all-features`; the dedicated job is the gated self-certification report producer.
+**CI:** the `rstix STIX interop self-certification` job in `.github/workflows/ci.yml` runs the full interop suite (every `TESTED` row), then `scripts/interop-report-gate.py` (report present, `generated_at` not older than `INTEROP_RUN_START`, Tables 55/56 cell content and traceability CSV rows match `gate-expectations.json`, every `TESTED` row passed, required features on), and uploads `target/interop-report/` as the `interop-report` artifact. The multi-OS `Test` matrix also executes the suite under `--all-features`; the dedicated job is the gated self-certification report producer.
+
+**Three-layer gate:**
+
+| Layer | Where | What it catches |
+| ----- | ----- | ---------------- |
+| 2 — export invariants | `tests/interop/harness/certification.rs` | Empty/wrong checklist `Result`, incomplete CSV, export before coverage passes |
+| 3 — manifest sync | `tests/interop/harness/gate_expectations.rs` + committed `gate-expectations.json` | Stale golden expectations after `manifest.toml` changes |
+| 1 — CI artifact gate | `scripts/interop-report-gate.py` | Stale/missing report, wrong table cells, CSV row/order/outcome drift, summary count mismatch |
+
+Regenerate expectations after manifest edits:
+
+```bash
+python3 scripts/generate-interop-gate-expectations.py
+```
 
 Local gate (same script CI uses):
 
--- a/crates/rstix/README.md
+++ b/crates/rstix/README.md
@@ -890,39 +890,45 @@ The tables above describe **implemented** behavior. Negative and rich fixtures f
 
 <a id="oasis-interop-test-suite"></a>
 
-### OASIS interop test suite
+### OASIS interop self-certification suite
 
-Test harness aligned with **STIX 2.1 Interoperability Version 1.0, Committee Specification Draft 01** (`stix-2.1-interop-v1.0-csd01`, 2021-10-23). It tracks manifest rows for all **21** use cases in that document; normative OASIS test-case JSON and per-requirement tests are added incrementally. The suite lives in `tests/interop/` (single Cargo target) with fixtures under `tests/fixtures/interop/`.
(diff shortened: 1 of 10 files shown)


scripts/interop-report-gate.py
```diff
@@ -0,0 +1,135 @@
...
+    if blocked != by.get("blocked"):
+        fail(
+            "blocked_rows mismatch: "
+            f"blocked_rows={blocked} disposition.blocked={by.get('blocked')}"
+        )
+    if report_only != by.get("report_only"):
+        fail(
+            "report_only_rows mismatch: "
+            f"report_only_rows={report_only} disposition.report_only={by.get('report_only')}"
+        )
```
mostafa: This block compares the report against itself, so it passes on reports that contain no evidence. Three payloads I ran against the script, each exiting 0:

1. Count fields absent. `{"generated_at": ..., "manifest_rows_by_disposition": {}, "harness_smoke_executed": 0, "features_enabled": {...}}` passes, printing `tested_passed=None`. `tested` and `tested_passed` are both `None`, and `None != None` is false, so the check is skipped rather than failed. Same for `blocked_rows` and `report_only_rows`. If the harness ever renames these fields, the gate goes quiet instead of red, which is the failure mode a gate exists to prevent.
2. Every row dropped. A report with `tested: 1` and `tested_rows_passed: 1` passes, because nothing pins how many rows there should be.
3. Every row demoted. `tested: 0`, `tested_rows_passed: 0`, `report_only: 513` passes. A self-certification gate satisfied by a report with zero verified requirements is the worst case of the three.

Two changes fix all three. Require the fields to be present and integral (`isinstance(value, int)` rather than `!=` against a possibly missing key), and pin absolute expectations rather than only relative ones. Floors read well here and only need touching when rows are deliberately added: `tested >= 506`, `blocked <= 2`, `report_only <= 5`, `manifest_rows_total == 513`. Cross-checking the counts against `manifest.toml` directly would be stronger still, since the manifest is the thing the report claims to summarize. There is precedent for pinned expectations in `fixture_walk.rs`, where `EXPECTED_GATING_FIXTURE_COUNT` and friends exist for exactly this reason.

scripts/interop-report-gate.py
```diff
@@ -0,0 +1,135 @@
...
+    except ValueError as err:
+        fail(f"INTEROP_RUN_START is not RFC 3339: {start_raw!r} ({err})")
+
+    if not REPORT_DIR.is_dir():
+        fail(f"missing report directory {REPORT_DIR}")
+
+    for name in REQUIRED_ARTIFACTS:
+        path = REPORT_DIR / name
+        if not path.is_file():
+            fail(f"missing required artifact {path}")
```
mostafa: The four non-JSON artifacts are only checked for existence. I truncated all of them to zero bytes and the gate still exited 0, so an empty Table 55, Table 56, traceability CSV, and risks file satisfy the check that is supposed to prove the report package was produced.

A non-empty check is one line each. Since the tables are the actual self-certification deliverable, asserting that Tables 55 and 56 each carry the expected 23 body rows would be better still and would catch a report generator that silently stops emitting rows.

mostafa (commented):
The workflow side is solid and the report claim checks out. The gate script is the part that needs work: it verifies the report is internally consistent but never that it contains anything, so several ways of losing the evidence entirely still exit 0. Details inline, with the exact payloads I ran.

What I verified:

- End to end works as advertised. `INTEROP_RUN_START=$(date -u ...)`, then the interop suite, then `python3 scripts/interop-report-gate.py` passes on a fresh report at 506 TESTED, 506 passed, 2 blocked, 5 report-only. Staleness and the pass-rate check both fail correctly when I feed them a report generated before `INTEROP_RUN_START` or with `tested_rows_passed` one short.
- The README claim holds against the artifacts. Tables 55 and 56 come out at 23 rows each, matching the 46 checklist cells the README cites, with no blank cell: 23 Pass on SXC, and 21 Pass plus the two expected `BLOCKED (unrepairable published test data)` rows for §3.12 and §3.16 on SXP. So "every Table 55/56 row is filled with a real result" is accurate as written, and the "not claimed" boundaries (no OASIS-issued certificate, no §4.1 product personas, no TAXII TXC) are the right ones to spell out.
- Workflow hygiene is clean: both actions pinned by full SHA with version comments, job-scoped `permissions: contents: read` with a comment, `persist-credentials: false`, `if-no-files-found: error`, and `INTEROP_RUN_START` recorded before the suite runs rather than after, which is what makes the stale check meaningful. `zizmor --pedantic` reports no findings on the modified workflow. `cargo fmt --all -- --check`, workspace clippy with all features, and clippy on the pinned 1.88.0 toolchain for the lean rstix feature set all pass.

One operational note that is yours rather than the contributor's: `ci.yml` has no aggregate job with a `needs:` list, so `rstix-interop` will not block a merge until it is added to the branch protection required checks. Until then it is advisory.

On the framing change: swapping "not a completed OASIS interoperability certification" for an explicit self-certification claim is defensible, since CSD01 self-certification is exactly a self-filled Table 55/56 and the tables are genuinely full. Worth being deliberate that the same edit drops the older hedge that a `TESTED` row means the harness recorded a pass rather than that the requirement was independently verified. The new text defines `Pass` honestly, so this is a judgment call rather than an inaccuracy, but the public claim got stronger in this PR while the underlying evidence did not change.

scripts/interop-report-gate.py
SecurityEnthusiast: Addressed in abb15c71.

- summary fields are now required integers (`require_int`); missing/renamed keys fail closed instead of comparing None to None
- absolute counts are pinned via committed `gate-expectations.json` (generated from `manifest.toml`): total 513, tested 506, blocked 2, report_only 5 — so a 1-row or all-report-only package fails
- Layer 3 keeps the golden in sync with the manifest; the interop suite fails if expectations drift

I went one step past floors: the gate also checks Table 55/56 cell content and full `traceability.csv` req_id/disposition/outcome against that same golden.

Open question to @mostafa in the commit body: is the committed JSON golden the right bar, or would you prefer a lighter summary-only pin without the artifact content compare?

scripts/interop-report-gate.py
SecurityEnthusiast: Addressed in abb15c71.

- every required artifact must be non-empty
- Tables 55/56 are parsed and must match the expected 23 body rows from `gate-expectations.json`, including Use case / Section / Verification / Result (so hollow or wrong-content tables fail, not just short ones)
- `traceability.csv` is parsed for header, row count/order, and expected outcomes (rejects MISSING/Fail)

Harness-side export invariants also panic before writing if checklist Results or CSV shape would be invalid after coverage.

SecurityEnthusiast:
> The workflow side is solid and the report claim checks out. The gate script is the part that needs work: it verifies the report is internally consistent but never that it contains anything, so several ways of losing the evidence entirely still exit 0. Details inline, with the exact payloads I ran.
> 
> What I verified:
> 
> * End to end works as advertised. `INTEROP_RUN_START=$(date -u ...)`, then the interop suite, then `python3 scripts/interop-report-gate.py` passes on a fresh report at 506 TESTED, 506 passed, 2 blocked, 5 report-only. Staleness and the pass-rate check both fail correctly when I feed them a report generated before `INTEROP_RUN_START` or with `tested_rows_passed` one short.
> * The README claim holds against the artifacts. Tables 55 and 56 come out at 23 rows each, matching the 46 checklist cells the README cites, with no blank cell: 23 Pass on SXC, and 21 Pass plus the two expected `BLOCKED (unrepairable published test data)` rows for §3.12 and §3.16 on SXP. So "every Table 55/56 row is filled with a real result" is accurate as written, and the "not claimed" boundaries (no OASIS-issued certificate, no §4.1 product personas, no TAXII TXC) are the right ones to spell out.
> * Workflow hygiene is clean: both actions pinned by full SHA with version comments, job-scoped `permissions: contents: read` with a comment, `persist-credentials: false`, `if-no-files-found: error`, and `INTEROP_RUN_START` recorded before the suite runs rather than after, which is what makes the stale check meaningful. `zizmor --pedantic` reports no findings on the modified workflow. `cargo fmt --all -- --check`, workspace clippy with all features, and clippy on the pinned 1.88.0 toolchain for the lean rstix feature set all pass.
> 
> One operational note that is yours rather than the contributor's: `ci.yml` has no aggregate job with a `needs:` list, so `rstix-interop` will not block a merge until it is added to the branch protection required checks. Until then it is advisory.
> 
> On the framing change: swapping "not a completed OASIS interoperability certification" for an explicit self-certification claim is defensible, since CSD01 self-certification is exactly a self-filled Table 55/56 and the tables are genuinely full. Worth being deliberate that the same edit drops the older hedge that a `TESTED` row means the harness recorded a pass rather than that the requirement was independently verified. The new text defines `Pass` honestly, so this is a judgment call rather than an inaccuracy, but the public claim got stronger in this PR while the underlying evidence did not change.

Thanks for the thorough check and the concrete failing payloads @mostafa. very helpful.

Gate holes you called out are fixed in abb15c71 (typed ints, absolute pins, non-empty artifacts, 23-row tables, plus content compare against a manifest-derived golden). Workflow hygiene notes taken as green.

On framing: kept the honest definition that Pass means the harness recorded a pass for that TESTED row. If you want the README hedge tightened further, say how you’d like it worded.

Curious for your take on the golden-JSON bar vs a lighter summary-only pin. see the question in abb15c71 please.

crates/rstix/README.md
```diff
@@ -942,30 +949,28 @@ cargo test -p rstix --test interop_sentinel --locked
...
-    path: target/interop-report/
+**CI:** the `rstix STIX interop self-certification` job in `.github/workflows/ci.yml` runs the full interop suite (every `TESTED` row), then `scripts/interop-report-gate.py` (report present, `generated_at` not older than `INTEROP_RUN_START`, Tables 55/56 cell content and traceability CSV rows match `gate-expectations.json`, every `TESTED` row passed, required features on), and uploads `target/interop-report/` as the `interop-report` artifact. The multi-OS `Test` matrix also executes the suite under `--all-features`; the dedicated job is the gated self-certification report producer.
+
+**Three-layer gate:**
+
+| Layer | Where | What it catches |
+| ----- | ----- | ---------------- |
+| 2 — export invariants | `tests/interop/harness/certification.rs` | Empty/wrong checklist `Result`, incomplete CSV, export before coverage passes |
+| 3 — manifest sync | `tests/interop/harness/gate_expectations.rs` + committed `gate-expectations.json` | Stale golden expectations after `manifest.toml` changes |
+| 1 — CI artifact gate | `scripts/interop-report-gate.py` | Stale/missing report, wrong table cells, CSV row/order/outcome drift, summary count mismatch |
```
mostafa: Cosmetic: the rows read 2, 3, 1. Ordering them 1, 2, 3 (or renaming to the order they fire in, which is export invariants, then manifest sync, then the CI artifact gate) would save the reader from reconciling the numbering against the sequence.

mostafa (commented):
Closed, and the approach is better than the floors I suggested. Pinning manifest-derived golden rows catches content drift, not just count drift, which is the thing that actually matters for a certification artifact.

I replayed the three payloads that used to exit 0 against the new script, and each now fails on the specific defect:

- Count fields absent: `summary.json field 'tested' must be an integer, got None`.
- Every row dropped: `manifest_rows_total mismatch: summary=1 expectations=513`.
- Every row demoted to REPORT_ONLY: `tested count mismatch: summary=0 expectations=506`.

I also tampered with a copy of a real report package to check the new content gates, and all four are caught: truncating `risks.md` to zero bytes, flipping one traceability outcome from `Pass` to `REPORT_ONLY`, deleting a single CSV row, and rewriting one Table 56 `Result` cell. The unmodified copy passes, so these are not false positives.

On the design, the part I checked hardest is whether the golden file could bless a regression. It cannot in the direction that matters: `generate-interop-gate-expectations.py` derives expectations from `manifest.toml` rather than from a previous report, CI never regenerates before gating, and `RSTIX_WRITE_GATE_EXPECTATIONS` is opt-in and unset in the workflow. The two implementations agree exactly too. Running the Python generator over the committed manifest reproduces `gate-expectations.json` byte for byte with an empty `git diff`, so the Rust and Python semantics have not drifted. Layer 3 does its job: I flipped one manifest row from `TESTED` to `REPORT_ONLY` without regenerating and the suite panicked with the regenerate instruction.

Worth stating plainly for whoever reads this later, since it is the one thing the gate cannot do: it protects against report and harness drift, not against a deliberate manifest demotion regenerated in the same commit. That path passes because both files change together, so it stays a review responsibility to read `manifest.toml` and `gate-expectations.json` diffs as a pair. That is the normal golden-file trade-off and I would not change anything for it.

Gates on the new head are clean: `cargo fmt --all -- --check`, workspace clippy with all features, clippy on the pinned 1.88.0 toolchain for the lean rstix feature set, `cargo test -p rstix --all-features --locked`, and `zizmor --pedantic` on the modified workflow with no findings. The end-to-end flow passes at 506 TESTED, 506 passed, 2 blocked, 5 report-only, and the README now describes what the gate checks accurately.

One cosmetic thing inline. My earlier note about the job needing to be in the branch protection required checks still stands, but that is on the maintainer side, not this PR.

crates/rstix/README.md
SecurityEnthusiast: Done.

mostafa (approved):

PR #451, opened 2026-08-07
test(rstix): TAXII 2.1 TXC interop self-certification

SecurityEnthusiast:
## Summary
- Add OASIS TAXII 2.1 Interoperability CSD01 (2022-03-30) **TXC** self-certification: all **39 Mandatory** §4.1 Table 51 rows (wiremock mock TXS + rustls mTLS for §3.1.3), **5 Optional** §3.13.2 rows as `REPORT_ONLY`.
- Align scenario depth to CSD01 Tables 2–50 (`User-Agent`, `WWW-Authenticate`, absolute/relative `api_roots`, collection id ordering, date-added headers, versions pagination via `added_after`, Status Table 28, envelope/Status `x_*`).
- Three-layer report gate (same pattern as STIX): harness export invariants, committed `gate-expectations.json` + generator, `scripts/taxii-interop-report-gate.py` row-level CI validation.
- Wire CI job `rstix TAXII TXC interop self-certification`.
- Bugfixes found while greening the suite:
  - Preserve `added_after` when clock skew is unset (`take()` footgun in `with_clock_skew_*`).
  - Pin rustls `ring` for §3.1.3 mTLS mock TXS when `ring` + `aws-lc-rs` are both linked.
  - Include `created_by_ref` identity in indicator parse fixture; assert filter queries via received requests; run scenarios in registry/section order (not lexicographic `test_id`).

## Test plan
- [x] `cargo clippy -p rstix --test taxii_interop --features taxii --locked -- -D warnings`
- [x] `cargo test -p rstix --all-features --locked` (incl. doctests)
- [x] `TAXII_INTEROP_RUN_START=…` + `cargo test -p rstix --test taxii_interop --features taxii --locked`
- [x] `python3 scripts/taxii-interop-report-gate.py`
- [ ] CI job uploads `taxii-interop-report` and gate fails on tampered Table 51 / CSV outcomes


crates/rstix/tests/taxii_interop/harness/mtls.rs
```diff
@@ -0,0 +1,123 @@
...
+    // `build_rustls_config` in the TAXII client).
+    let provider = rustls::crypto::ring::default_provider();
+    let _ = provider.install_default();
+}
+
+/// Serve one HTTPS discovery response that requires a client certificate (CSD01 §3.1.3).
+pub async fn certificate_auth() {
+    install_ring_provider();
+
+    let ca_pem = read_pem("ca.pem");
```
mostafa: This is the blocker. `live_cert_dir()` points at `tests/taxii-live/fixtures/certs`, which is not in the repo. It only exists after someone runs the live harness `generate-certs.sh`, which needs the OpenSSL CLI and, as a side effect, rewrites the tracked `tests/taxii-live/coredns/taxii.test.zone`. Nothing in CI runs it, and nothing in the README or manifest mentions it as a prerequisite, so the green local run depends on leftover untracked state from a previous live-harness session.

Suggested fix: mint the CA, server and client material in process with `rcgen`, which is already a dev-dependency in `rsigma-cli` and `rsigma-runtime` and already in `Cargo.lock`. That keeps the row hermetic and keeps this suite independent of the docker harness. Adding a `generate-certs.sh` step to the CI job would work too, but it couples a gated job to a manual setup script and still leaves a fresh clone broken locally. Skipping the row when the certs are absent is the one option I would avoid, since §3.1.3 is Mandatory and a silent skip is exactly what the three-layer gate exists to prevent.

crates/rstix/src/taxii/client.rs
```diff
@@ -949,8 +954,41 @@ fn with_clock_skew_versions_filter(
...
     adjusted
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::core::TaxiiTimestamp;
+
+    #[test]
+    fn clock_skew_helper_preserves_added_after_when_skew_absent() {
```
mostafa, line 980: The two sibling helpers get the identical fix but no coverage. `with_clock_skew_object_filter` and `with_clock_skew_versions_filter` had the same `take()` footgun, so the same pair of assertions against each would keep all three honest rather than only the one.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to RSigma are documented in this file. Each entry correspond
 
 ## [Unreleased]
 
+### rstix: TAXII 2.1 TXC interop self-certification (C-1) (#451)
```
mostafa, line 7: Two user-visible changes are missing from the entry.

`parse_www_authenticate` is public API (re-exported from `taxii::mod`), and its results change for any multi-challenge header: auth-params no longer split into fake schemes. Anyone matching on `challenges.len()` or on `scheme` sees different values after this.

`TaxiiEnvelope` also gains a public `custom` field, so struct-literal construction of it no longer compiles. `TaxiiStatus` and `ManifestResponse` already carried the field, so the shape is consistent, but 0.21.0 is published and this is worth a line.

crates/rstix/src/taxii/www_authenticate.rs
```diff
@@ -33,19 +33,35 @@ pub fn parse_www_authenticate(value: &str) -> Vec<AuthChallenge> {
...
+        let ch = chars[i];
         match ch {
             '"' => {
                 in_quotes = !in_quotes;
                 current.push(ch);
+                i += 1;
             }
             ',' if !in_quotes => {
-                parts.push(std::mem::take(&mut current));
+                let rest: String = chars[i + 1..].iter().collect();
```
mostafa: Minor, on untrusted input: this collects the remainder of the header into a fresh `String` at every unquoted comma, so a header with many commas is quadratic. Scanning forward in place over `chars[i + 1..]` avoids the allocation and reads about the same.

While here, `in_quotes` toggles on every `"` including escaped ones, so a value like `realm="a\",Basic b"` desyncs the parity. The new `starts_new_challenge` guard rescues it in practice because the following token contains a quote and fails the charset check, which is a nice property worth pinning with a test.

crates/rstix/README.md
```diff
@@ -629,13 +629,53 @@ Notes:
...
+| Mock TXS | wiremock + local rustls mTLS acceptor (§3.1.3) | OASIS-shaped responses; no network; rustls `ring` provider pinned (avoids dual `ring`/`aws-lc-rs` panic) |
+| Report | `target/taxii-interop-report/` | `txc-table-51.md`, `traceability.csv`, `summary.json`, `risks.md` |
+| CI gate | `scripts/taxii-interop-report-gate.py` | Content + freshness (`TAXII_INTEROP_RUN_START`) |
+
+**Claim boundaries:** every Mandatory Table 51 cell is filled from automated results (`Pass`). The five Optional §3.13.2 additional-filter rows are `REPORT_ONLY`. This is **self-certification evidence**, not an OASIS-issued certificate.
+
+**Assertion depth:** scenarios follow CSD01 Tables 2–50 (from `plan/taxii-2.1-interop-v1.0.docx`), not only checklist titles — including `User-Agent` (§2.1.4), `WWW-Authenticate` challenges (Table 2), absolute+relative `api_roots`, collections ordered by `id`, `X-TAXII-Date-Added-*` headers, versions pagination via `added_after` from `X-TAXII-Date-Added-Last` (Tables 48–49), Status Table 28 fields, and envelope/Status `x_*` custom properties (Table 50 / `TaxiiEnvelope::with_custom`).
```
mostafa: `plan/taxii-2.1-interop-v1.0.docx` is not tracked, so this path is a dead reference for readers, and the same string is emitted into the published `risks.md` by `certification.rs:326`. The CSD01 title and date are already stated just above, so citing the document by name alone reads better in both places.

crates/rstix/README.md
```diff
@@ -629,13 +629,53 @@ Notes:
...
+| Component | Path | Role |
+| --------- | ---- | ---- |
+| Suite | `tests/taxii_interop/` | Custom runner (`harness = false`); 39 Mandatory scenarios |
+| Manifest | `tests/fixtures/taxii-interop/manifest.toml` | Table 51 `req_id` → `test_id` → disposition |
+| Gate expectations | `tests/fixtures/taxii-interop/gate-expectations.json` | Manifest-derived golden rows for Table 51 and traceability CSV (regenerate with `scripts/generate-taxii-interop-gate-expectations.py` when the manifest changes) |
+| Mock TXS | wiremock + local rustls mTLS acceptor (§3.1.3) | OASIS-shaped responses; no network; rustls `ring` provider pinned (avoids dual `ring`/`aws-lc-rs` panic) |
+| Report | `target/taxii-interop-report/` | `txc-table-51.md`, `traceability.csv`, `summary.json`, `risks.md` |
+| CI gate | `scripts/taxii-interop-report-gate.py` | Content + freshness (`TAXII_INTEROP_RUN_START`) |
+
+**Claim boundaries:** every Mandatory Table 51 cell is filled from automated results (`Pass`). The five Optional §3.13.2 additional-filter rows are `REPORT_ONLY`. This is **self-certification evidence**, not an OASIS-issued certificate.
```
mostafa: One sentence would sharpen this. The counterparty here is a mock authored in this repo from the CSD01 tables, so the evidence bounds client behavior against our reading of the spec rather than against an independent TXS. That is a real step down from the STIX suite, where every fixture is published bytes with a provenance sidecar recording repairs and divergences. Saying so explicitly makes the claim harder to over-read later, and it costs nothing given `risks.md` already notes the stand-in.

mostafa (changes requested):
The two library fixes are real and both are genuinely guarded by scenarios, and the three-layer gate carries the STIX hardening over intact. One blocker: the suite cannot run on a clean checkout.

**Blocker.** `tc_3_1_3_certificate_auth` reads PEMs from `crates/rstix/tests/taxii-live/fixtures/certs/`, which is not tracked. That directory is output of the opt-in live harness (`tests/taxii-live/generate-certs.sh`, needs the OpenSSL CLI). On a fresh clone the suite panics at `mtls.rs:25` before any scenario runs:

```
thread 'main' panicked at crates/rstix/tests/taxii_interop/harness/mtls.rs:25:47:
read .../tests/taxii-live/fixtures/certs/ca.pem: No such file or directory (os error 2)
```

That takes down both the new job and `cargo test --workspace --all-features --locked` (ci.yml:272), since `taxii_interop` is a `--all-features` target. The README runbook added in this PR fails the same way. Details and a suggested fix inline.

**Verified locally** (after generating the certs by hand, everything else is green):

* Gate rejects all seven fail-open probes I threw at it: absent count fields, rows dropped to 1, all 39 demoted to `REPORT_ONLY`, zero-byte `risks.md`, flipped CSV outcome, deleted CSV row, altered Table 51 `Result` cell. That covers the unchecked box in your test plan for the gate half.
* Layer 2 catches a manifest disposition edit without regeneration (`gate-expectations.json is stale`), and `generate-taxii-interop-gate-expectations.py` reproduces the committed file byte for byte from `manifest.toml`, so the golden file is manifest-derived and not report-derived.
* Mutation probes: reverting the `added_after` skew fix fails `filters.rs:52`, reverting the `WWW-Authenticate` split fix fails `auth.rs:26`. Both fixes have real regression protection rather than only unit coverage.
* 39 scenario functions for 39 Mandatory rows, one to one, so no row borrows another row's evidence.
* `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo +1.88.0 check --workspace --all-features`, `cargo test -p rstix --all-features`, `zizmor --pedantic .github/workflows/ci.yml`: all clean. Gate reports 39 tested passed, 5 report only, 44 rows.

`risks.md` is honest about the mock stand-in and the persona boundary, which is the right call.

mostafa (approved):
All six items addressed. Re-verified at e1ff9a7.

**Blocker cleared.** I deleted the `taxii-live/fixtures` tree entirely and ran the suite from a checkout with no cert material anywhere: 39 tested passed, 5 report only, gate ok. `mint_mtls_material` gives the server cert a `localhost` SAN plus a `127.0.0.1` IP SAN and splits `ServerAuth` and `ClientAuth` EKUs across the two leaves, so the row exercises real chain and client-cert verification rather than a shortcut. Nothing in the suite touches the live harness now.

On the `rcgen` backend choice: I checked whether `aws_lc_rs` pulls a second crypto backend into this test binary, and `aws-lc-sys` was already in the graph on `main` via `rustls` and `reqwest`, so this adds no build cost and no new provider ambiguity beyond what `install_ring_provider` already pins. Nothing to change.

**The rest.** All three sibling clock-skew paths are now covered on both branches through the shared `assert_added_after_query`. The challenge splitter scans `&chars[i + 1..]` in place, so the per-comma allocation is gone, and the escaped-pair arm is a real fix rather than leaning on the token guard. Worth noting that the new test earns its keep in a non-obvious way: I simulated the pre-commit splitter on `Basic realm="a\",Basic b", Bearer realm="x"` and it also returns two challenges, so the length assertion alone would pass. It is the `scheme` and `params` assertions that catch it, since the old parity desync yields `realm="a\"` for the first challenge and `Basic` for the second.

The changelog now calls out both API changes under an explicit breaking-vs-0.21.0 heading, and the claim-boundary sentence landed in both the README and the generated `risks.md`, which is where it does the most good since the report is the artifact that outlives the PR. The dead `plan/` path is gone from both.

**Gates:** `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test -p rstix --all-features` (including the interop target and doctests), `cargo +1.88.0 check --workspace --all-features`, and the interop suite plus `taxii-interop-report-gate.py` on a cert-free tree: all clean.

PR #499, opened 2026-09-20
feat(rstix): validate-on-ingest for TAXII pages

SecurityEnthusiast:
## Summary
- Validate each TAXII page as a synthetic bundle before `StixStore::import_objects`.
- `IngestOptions::reject_invalid_pages` defaults to `true`; failures recorded in `IngestValidationReport`.
- Tests and README/CHANGELOG updates for validate-on-ingest.
## Test plan
- [*] `cargo test -p rstix --features taxii,store,validate --locked -- taxii_store`
- [*] `cargo test -p rstix --features taxii,store,validate --locked -- ingest_validate`
- [*] Full workspace gate (`fmt`, `clippy`, `cargo test --workspace --all-features --locked`)


crates/rstix/src/taxii/ingest.rs
```diff
@@ -13,6 +18,104 @@ use super::{TaxiiClient, TaxiiError, TaxiiFilter};
...
+    }
+}
+
+#[cfg(feature = "validate")]
+impl IngestOptions {
+    /// Validate each page with [`Validator::interop_strict`] and reject invalid pages.
+    pub fn interop_strict() -> Self {
+        Self {
+            validator: Some(Validator::interop_strict()),
+            reject_invalid_pages: true,
```
mostafa: `interop_strict()` is the wrong named constructor for paginated ingest. It runs all twelve phases including References at `Leniency::Zero`, so `STIX-W0010` (“reference does not resolve within the bundle”) fails the page. TAXII pages are not closed bundles; the deferred `audit_unresolved_refs` after all pages exists specifically so forward refs across pages resolve. I ran the existing two-page report→indicator fixture through this constructor: page 1 is skipped, page 2 is imported, `objects_added=1`.

Use `Validator::producer_strict()` (skips References) or `validate_object` per object as the ingest default. If `interop_strict` stays, it needs a loud “closed-bundle pages only” warning and must not be the README example.

crates/rstix/src/taxii/ingest.rs
```diff
@@ -91,23 +206,54 @@ pub async fn ingest_collection_with_bundle_id(
...
+
+    #[cfg(not(feature = "validate"))]
+    let _ = &bundle_id;
+
+    #[cfg(feature = "validate")]
+    if let Some(validator) = &options.validator {
+        report.validation.pages_validated += 1;
+        let bundle = Bundle::from_objects(bundle_id.clone(), envelope.objects.clone());
+        let validation = validator.validate_bundle(&bundle);
+        if !validation.is_valid() {
```
mostafa: Whole-page reject: `!validation.is_valid()` drops every object on the page. A TAXII page is a pagination window, so one SHOULD (the CVE-prefix `W0010` in the new test) or one cross-page ref sinks otherwise-valid neighbors. Prefer per-object validation (`Validator::validate_object`) with reject/import decided per object, or document and test the all-or-nothing choice explicitly.

Also, `Bundle::from_objects(bundle_id.clone(), envelope.objects.clone())` copies the whole page only to walk it. `validate_object` over `&envelope.objects` avoids that clone.

When `validate` is off this function has unused `options` and `page_index` (and the `Bundle` import at line 8 is unused). `let _ = (options, page_index);` next to the existing `let _ = &bundle_id` keeps `taxii-store`-only clippy green.

crates/rstix/src/taxii/ingest.rs
```diff
@@ -13,6 +18,104 @@ use super::{TaxiiClient, TaxiiError, TaxiiFilter};
...
+    pub rejected_object_count: usize,
+    /// Per-page failure details.
+    pub failures: Vec<IngestValidationFailure>,
+}
+
+#[cfg(feature = "validate")]
+impl IngestValidationReport {
+    /// True when no page was rejected.
+    pub fn is_valid(&self) -> bool {
+        self.pages_rejected == 0
```
mostafa: `is_valid()` is `pages_rejected == 0`. Under `allow_invalid_pages()`, failures are recorded but `pages_rejected` stays 0, so `report.validation.is_valid()` is true while `failures` is non-empty. Either key it off `failures.is_empty()`, or rename (e.g. `no_pages_rejected`) so the README `assert!(report.validation.is_valid())` cannot be misread.

crates/rstix/README.md
```diff
@@ -619,15 +619,40 @@ use rstix::taxii::{ingest_collection, TaxiiClient, TaxiiClientConfig, TaxiiFilte
...
+use rstix::taxii::{IngestOptions, ingest_collection_with_bundle_id};
+
+let report = ingest_collection_with_bundle_id(
+    &client,
+    &store,
+    api_root_url,
+    "col1",
+    TaxiiFilter::new(),
+    bundle_id,
+    IngestOptions::interop_strict(),
```
mostafa: This example is the path that drops cross-page refs (see ingest.rs). The note two bullets below still says reference checks run after all pages against the full store, which is only true for `IngestOptions::default()` (`validator: None`). Those two claims cannot both stand.

Also, “Invalid pages are skipped by default (`reject_invalid_pages: true`)” is easy to over-read: `ingest_collection` does not validate at all. `reject_invalid_pages` only matters once a validator is attached. Spell that out, and prefer `producer_strict` / per-object validation in the sample.

CHANGELOG.md
```diff
@@ -4,7 +4,11 @@ All notable changes to RSigma are documented in this file. Each entry correspond
 
 ## [Unreleased]
 
-### Security: rustls advisory + cargo-deny config
+### rstix: validate-on-ingest for TAXII collections (#499)
+
+`ingest_collection` now returns [`IngestReport`](crates/rstix/README.md) (`import` + optional `validation`). With the `validate` feature, [`IngestOptions::interop_strict()`](crates/rstix/README.md) validates each TAXII page as a synthetic bundle before store import and rejects invalid pages by default.
```
mostafa: Call the break vs 0.22.0 the way #451 did: `ingest_collection` / `ingest_collection_with_bundle_id` now return `IngestReport` instead of `ImportReport`, and `ingest_collection_with_bundle_id` takes a new `IngestOptions` argument. Callers matching the old signature will not compile.

The `[IngestReport](crates/rstix/README.md)` links go to the README, not the types. Plain backticks are enough here.

`docs/content/library/rstix.md` still describes ingest as `ingest_collection` with no `IngestReport` / `IngestOptions`; that page needs the same update.

crates/rstix/tests/taxii/ingest_validate_tests.rs
```diff
@@ -0,0 +1,108 @@
...
+        "name": "Bad CVE ref",
+        "external_references": [{
+            "source_name": "cve",
+            "external_id": "2016-1234"
+        }]
+    })
+}
+
+#[tokio::test]
+async fn ingest_rejects_invalid_page_under_interop_strict() {
```
mostafa: This is the only reject test, and it uses a CVE `external_id` SHOULD (`2016-1234` → `STIX-W0010`), which only fails because `interop_strict` promotes warnings. It would pass under `producer_strict` / Standard leniency, so it does not pin “invalid STIX is not stored.”

Please add:
1. A MUST-level invalid object (missing required field or bad id) that is rejected under a non-Zero profile.
2. The existing two-page forward-ref fixture under the ingest validator, asserting both objects are imported (this is the case that currently fails with `interop_strict`).
3. `allow_invalid_pages()` still imports and still records `failures`.
4. A mixed page (one bad object + one valid neighbor) once reject granularity is decided.

mostafa (changes requested):
The plumbing (options, report, feature gates, skip-on-fail) is in the right shape, but the advertised default profile is unsafe for paginated TAXII ingest.

**Blocker.** `IngestOptions::interop_strict()` validates each TAXII page as a synthetic `Bundle` with `Validator::interop_strict()`, which runs the References phase at `Leniency::Zero`. Unresolved refs are `STIX-W0010` warnings, and Zero leniency treats warnings as failures, so a spec-legal object whose target is on a later page is dropped before import. I replayed the existing `ingest_collection_resolves_forward_refs_across_pages` fixture through `interop_strict()`: page 1 (report → indicator on page 2) is rejected with `reference \`indicator--…\` does not resolve within the bundle`, and only the page-2 indicator is stored (`objects_added=1`). That undoes the two-phase ingest work from #387, whose whole point was that TAXII pages are not closed bundles. The README still says reference checks run after all pages against the full store, which is only true for `audit_unresolved_refs` when no validator is set.

`producer_strict()` already skips References, and `Validator::validate_object` exists for per-object checks with no bundle resolution. Either of those is the right ingest default. Keep `interop_strict` only if you document it as “this page is a closed bundle” and do not recommend it for collection ingest.

Related: one diagnostic on a page rejects every neighbor on that page. TAXII pages are pagination, not semantic groups, so a single CVE-prefix SHOULD (`W0010` again) drops otherwise-valid objects. Object-level reject, or at least a test that pins the all-or-nothing choice, is needed before this ships.

**Also must-fix**

* `cargo clippy -p rstix --features taxii-store -- -D warnings` fails on this PR: unused `Bundle` import, unused `options` / `page_index` in `import_page`, and `clippy::derivable_impls` on `IngestOptions::Default` (only empty when `validate` is off). Workspace `--all-features` hides it; the `taxii-store` test target in the PR test plan does not.
* Breaking vs 0.22.0: `ingest_collection` / `ingest_collection_with_bundle_id` now return `IngestReport`, and the latter takes a new `IngestOptions` argument. CHANGELOG mentions the return type, not the extra argument.
* `docs/content/library/rstix.md` is unchanged. Public ingest API belongs on the docmd site.
* The only reject test uses a CVE `external_id` SHOULD (`2016-1234` → `W0010`), which only fails because Zero leniency promotes warnings. There is no MUST-level invalid object, no `allow_invalid_pages` path, no mixed valid/invalid page, and no cross-page case under a validator.

Existing tests under default options still pass, fmt is clean, and clippy is clean with `taxii-store,validate`. The feature split itself is fine: `ingest_collection` does not validate unless a validator is attached.

SecurityEnthusiast:
Thanks for the detailed review @mostafa .
## Blocker: `interop_strict()` on paginated ingest
- Removed whole-page `validate_bundle` + `interop_strict()` as the ingest path.
- Added `IngestOptions::producer_strict()` as the recommended profile for collection ingest.
- Each object is validated as a one-object synthetic `Bundle` with `validate_bundle` under the chosen profile. `producer_strict` skips References, so forward refs across TAXII pages still import and resolve after the full run via `audit_unresolved_refs`.
- Kept `interop_strict()` with docs that it is zero-leniency per-object (warnings like `STIX-W0010` fail), not for ordinary paginated ingest.
Note on `validate_object`: I tried per-object `Validator::validate_object` first. TypeDiscrimination expects a bundle root (`E0002` on `"type": "indicator"`), so every object failed validation. One-object synthetic bundles match the pipeline wire shape without reintroducing closed-bundle page semantics.
## Whole-page reject → per-object reject
- Reject granularity is per object, not per page.
- Renamed for clarity: `reject_invalid_objects` / `allow_invalid_objects()`; report fields `objects_validated`, `objects_rejected`; failures include `object_id`.
## Clippy (`taxii-store` without `validate`)
- `Bundle` import is `#[cfg(feature = "validate")]` only.
- `let _ = (&bundle_id, options, page_index)` when `validate` is off.
- `cargo clippy -p rstix --all-targets --features taxii-store -- -D warnings` is clean.
## Breaking vs 0.22.0 + docs
- CHANGELOG: `IngestReport` return type + new `IngestOptions` arg on `ingest_collection_with_bundle_id`; `(#499)` in heading.
- `docs/content/library/rstix.md`: new Collection ingest (`taxii-store`) section with breaking note and `producer_strict` example.
- README: default ingest does not validate; `reject_invalid_objects` only applies when a validator is attached; example uses `producer_strict()`.
## Tests added/updated
1. MUST-level reject under `producer_strict`: identity with short timestamp (`STIX-E0012`) — parses on the wire, fails validation (not parse).
2. Forward-ref two-page fixture under `producer_strict`: both objects imported, `unresolved_references` empty.
3. `allow_invalid_objects()`: imports + records `failures`.
4. Mixed page: invalid neighbor rejected, valid indicator on same page imported.
5. `interop_strict`: CVE-prefix SHOULD (`STIX-W0010`) still rejected at zero leniency.
## Local verification
```bash
cargo fmt --all -- --check
cargo clippy -p rstix --all-targets --features taxii-store --locked -- -D warnings
cargo clippy -p rstix --all-targets --features taxii-store,validate --locked -- -D warnings
cargo test -p rstix --features taxii-store,validate --locked --test taxii_store
```
Please re-review when CI is green. Happy to adjust naming (`reject_invalid_objects` vs `reject_invalid_pages`) if you prefer minimal API churn on this PR.

crates/rstix/src/taxii/ingest.rs
```diff
@@ -13,6 +19,118 @@ use super::{TaxiiClient, TaxiiError, TaxiiFilter};
...
+            validator: Some(Validator::producer_strict()),
+            reject_invalid_objects: true,
+        }
+    }
+
+    /// Validate each object with [`Validator::interop_strict`] at zero leniency.
+    ///
+    /// **Stricter than [`Self::producer_strict`]:** warnings (for example `STIX-W0010`) fail
+    /// validation. Still skips References (per-object validation). Prefer [`Self::producer_strict`]
+    /// for paginated collection ingest unless you need interop-grade strictness on each object.
```
mostafa: `Validator::interop_strict()` includes References. A one-object synthetic bundle does not skip that phase: a report whose `object_refs` target is not in that one-object wrapper still fails `STIX-W0010` and is not imported. I confirmed `added=0 rejected=1` on the same forward-ref report the `producer_strict` test now keeps.

Please delete “Still skips References (per-object validation).” Say that this profile is zero-leniency **and** still treats each object as a closed bundle for refs, which is why `producer_strict` is the ingest default. The README sentence that claims References is skipped whenever a validator is attached (around the validate-on-ingest intro) needs the same narrowing.

mostafa (approved):
The original review items are addressed. Re-verified at b41b9f55.

**Blocker cleared.** Ingest default is now `IngestOptions::producer_strict()`, reject is per object, and the two-page report→indicator fixture imports both objects with empty `unresolved_references`. Mixed-page, MUST-level (`E0012` short timestamp), `allow_invalid_objects`, and zero-leniency CVE SHOULD cases are all present. CHANGELOG names both breaks vs 0.22.0, and `docs/content/library/rstix.md` has the ingest section.

The `validate_object` → `E0002` note is a real pipeline limitation. One-object synthetic bundles are a fair workaround for TypeDiscrimination as long as the profile skips References (`producer_strict`).

**One leftover, not a re-block.** `IngestOptions::interop_strict()` still runs the References phase. I replayed a lone report with an outbound `object_refs` through it: `added=0 rejected=1`, diagnostic `STIX-W0010 reference … does not resolve within the bundle`. The constructor doc (and README line 627) say References is skipped for any attached validator / for `interop_strict`. That is only true for `producer_strict`. Please drop the “still skips References” sentence and say instead that `interop_strict` is zero-leniency **and** still closed-bundle on refs, which is why it is not for paginated ingest. Details inline.

**Gates:** `cargo fmt --check`, `cargo clippy -p rstix --all-targets --features taxii-store,validate -- -D warnings`, `cargo test -p rstix --features taxii-store,validate --test taxii_store`: all clean. `taxii-store` without `validate` still fails clippy on three pre-existing `collapsible_if` lints in `model/validation.rs`, not this PR.

mostafa (approved):
@SecurityEnthusiast Thank you for your contribution. LGTM! 🚀

